PR #1, opened 2021-03-10
Compilation fix

sergey-unikie:
`colcon build` command fails in `fog_sw` repo. This little fix make it pass


joenpera:
I will integrate this fix in.  This needs also changes to mesh_com subscriber and systemd mesh.service (using absolute paths currently to mesh.sh).

PR #7, opened 2021-04-23
Camp

joenpera:
changing QoS policy to follow configuration defined in fast RTPS profile XML (defined in https://github.com/tiiuae/fogsw_configs)


pemanty (approved):
approved

PR #9, opened 2021-04-26
Persist AP creation in sc-mesh-secure-deployment module

mbaddeley:
Creates a mesh-ap.sh script run by an ap@.service so it persists after reboot. 
- Requested IP is passed in as an instance to the ap services (i.e. 60 for 60.0.0.1).
- wpa_supplicant instance is created to configure the interface at boot.
- Interface is also saved in an ap.conf file andto create the iptables entries in mesh-ap.sh script.
- Added a "Remove AP" option to the configuration to be able to remove the AP.


modules/sc-mesh-secure-deployment/configure.sh
```diff
@@ -105,19 +144,18 @@ function server {
...
-  # Make the client
+  # Make the server
   pushd .
   cd ../..
-  make mesh_tb_client
+  make client
   popd
   # Connect to the same AP as the server
-  read -p "> We MUST be on the same network as the Server. Connect to AP? (Y/N): " confirm
+  read -p "> We need to be connect to the same network as the server... Connect to an Access Point? (Y/N): " confirm
```
martin-tii, line 153: as a suggestion: we should see if the interface is already connected to an access point with `iwconfig ` to avoid asking again to reconnect

martin-tii (approved):
LGTM

common/scripts/example.sh
```diff
@@ -0,0 +1,39 @@
...
+iptables --delete-chain
+
+# Delete all chains that are not in default filter and nat table
+
+iptables --table nat --delete-chain
+
+# Set up IP FORWARDing and Masquerading
+# eno1 is "private" wlx is "public"
+
+iptables --table nat --append POSTROUTING --out-interface wlx00e04c6770ed -j MASQUERADE
```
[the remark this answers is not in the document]
mbaddeley: This was just a temp file, should'nt have been added. Removing.

modules/sc-mesh-secure-deployment/conf/ap.conf
```diff
@@ -0,0 +1,7 @@
+  network={
+    ssid="AP-TEST"
+    mode=2
+    key_mgmt=WPA-PSK
+    psk="ssrcpassword"
```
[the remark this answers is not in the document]
mbaddeley: Good idea, will create an issue for this and we can pick it up later

modules/sc-mesh-secure-deployment/configure.sh
mbaddeley: Good idea, will make this change and push.

modules/sc-mesh-secure-deployment/configure.sh
```diff
@@ -58,44 +82,59 @@ cat <<EOF > tools/wpa_tools/access_point.conf
     frequency=2437
 }
 EOF
-cd tools/wpa_tools
-chmod +x create_access_point.sh
-sudo bash create_access_point.sh $choice $ip
-cd ../..
+# FIXME: Each time you echo this it's gonna add to the end of the files and
+#        create duplicate lines
+echo "INTERFACES=$1" >> /etc/default/isc-dhcp-server
```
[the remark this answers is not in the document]
mbaddeley: It may contain multiple interfaces so probably a bad idea to overwrite. I like the backup and recovery though, this could be a way forward. Will create an issue.

common/scripts/mesh-ap.sh
```diff
@@ -0,0 +1,15 @@
...
+  option routers $2.10.0.1;
+}
+EOF
+
+# give configuration to the passed interface
+ifconfig "$1" $2.10.0.1
+service isc-dhcp-server start
+sysctl -w net.ipv4.ip_forward=1
+# output all traffic from $2.10.0.0/24 to bat0 (i.e. mesh inf with a gw route)
+sudo iptables -t nat -A POSTROUTING -s $2.10.0.0/24 -o bat0 -j MASQUERADE
```
joenpera, line 15: This is not creating AP as file name says.  This is screating DHCP and routing.  Should it be renamed?

modules/sc-mesh-secure-deployment/configure.sh
mbaddeley: Ah we need to actually figure out a better way of doing this as the config is needed for the 1st node joined (i.e. gateway). Will create an issue.

common/scripts/mesh-ap.sh
mbaddeley: After thinking about this - yes it doesn't specifically create the AP, but it's used WHEN creating the AP. Otherwise it's name is mesh-dhcp-routing.sh, which I think is more confusing?

PR #17, opened 2021-05-24
11s support added to node, +some refactoring

joenpera:
- 11s support added to mesh_com ros node
- mesh-11s.sh updated to same level as ibss script (modes: mesh, ap, off)
- mesh_com python code refactoring
- api_version 1 for ibss and api_version 2 for 11s
- version update to 0.3.0

### DIFF
--- a/common/scripts/mesh-11s.sh
+++ b/common/scripts/mesh-11s.sh
@@ -3,27 +3,47 @@
 function help
 {
     echo
-    echo "Usage: sudo ./mesh_11s.sh <mode> <ip> <mask> <AP MAC> <key> <essid> <freq> <txpower> <country>"
+    echo "Usage: sudo ./mesh_11s.sh <mode> <ip> <mask> <AP MAC> <key> <essid> <freq> <txpower> <country> [<interface>] [<phyname>]"
+    echo "Parameters:"
+    echo "	<mode>"
+    echo "	<ip>"
+    echo "	<mask>"
+    echo "	<AP MAC>"
+    echo "	<WEP key>"
+    echo "	<essid>"
+    echo "	<freq>"
+    echo "	<txpower>"
+    echo "	<country>"
+    echo "	<interface>" - optional
+    echo "	<phyname>"   - optional
     echo
     echo "example:"
-    echo "sudo mesh.sh mesh 192.168.1.2 255.255.255.0 00:11:22:33:44:55 1234567890 mymesh2 5220 30 fi"
+    echo "sudo ./mesh.sh mesh 192.168.1.2 255.255.255.0 00:11:22:33:44:55 1234567890 mymesh2 5220 30 fi wlan1 phy1"
     echo "sudo mesh.sh ap"
     exit
 }
 
-echo "Solving wifi device name.."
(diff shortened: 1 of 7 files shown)


modules/mesh_com/package.xml
```diff
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>mesh_com</name>
-  <version>0.2.0</version>
+  <version>0.3.0</version>
   <description>ROS Mesh Node</description>
   <maintainer email="mika.joenpera@unikie.com">joenpera</maintainer>
   <license>TODO: License declaration</license>
```
mbaddeley, line 8: We should probably have a discussion about this on Thursday?

mbaddeley, line 8: We should probably have a discussion about this (what license) on Thursday? BSD?

common/scripts/mesh-11s.sh
```diff
@@ -82,32 +109,116 @@ EOF
...
+        echo "DRONE_DEVICE_ID not available"
+
+cat <<EOF >/var/run/wpa_supplicant-ap.conf
+ctrl_interface=DIR=/var/run/wpa_supplicant
+ap_scan=1
+p2p_disabled=1
+network={
+    ssid="debug-hotspot"
+    mode=2
+    frequency=5220
```
mbaddeley, line 143: We should set the freq with $7 (<freq>)

common/scripts/mesh-11s.sh
```diff
@@ -82,32 +109,116 @@ EOF
...
+    key_mgmt=WPA-PSK
```
mbaddeley, line 144: It would be nice to be able to pass a param to turn this off/on for testing purposes. For example, we may actually wish to have this unsecured for easy on-air monitoring, or to improve performance. Doesn't need to be fixed here but let's raise it as a github issue?

common/scripts/mesh-11s.sh
```diff
@@ -82,32 +109,116 @@ EOF
...
+    psk="1234567890"
```
mbaddeley, line 145: Again not an issue with this PR, but just in general this doesn't seem very secure :P We should create a github issue to address this (hashing it?).

common/scripts/mesh-11s.sh
```diff
@@ -82,32 +109,116 @@ EOF
...
-      nmcli radio wifi on
-
       # find ap parameters from enclave
       if [ -z "$DRONE_DEVICE_ID" ]
       then
-            echo "DRONE_DEVICE_ID not available"
-            nmcli device wifi hotspot con-name debug-hotspot ssid debug_hotspot band a channel 36 password 1234567890
+        echo "DRONE_DEVICE_ID not available"
+
+cat <<EOF >/var/run/wpa_supplicant-ap.conf
```
mbaddeley, line 136: This is a general comment for creating aps (in both mesh-11s and mesh-ibss). At the moment these fields are all hardcoded, and we create aps slightly differently in sc-secure-deployment. We should create a git issue to fix and unify this.

PR #18, opened 2021-05-24
Martin dev

martin-tii:
Mutual Authentication. 
The server verifies the client certificate, if it's valid, he sends the mesh information along with his certificate.
The node verifies the server certificate, if it's valid, the node sends the MAC address of the Mesh as an Acknowledgment.
The Mesh MAC is printed on the web page of the authenticated/blocked nodes.


modules/sc-mesh-secure-deployment/src/server-mesh.py
```diff
@@ -87,12 +87,25 @@ def add_message(uuid):
...
+    new_md5 = hashlib.md5(new).hexdigest()
+    return old_md5 == new_md5
+
 
 def add_default_route(ip_gateway):
     inter = netifaces.interfaces()
     for interf in inter:
         # TODO: what it if doesn't start with wlan???
-        if interf.startswith('wlan'):
+        #if interf.startswith('wlan'):
```
mbaddeley: So I had also thought that we should do this, but does this then not introduce a problem if there's two interfaces that begin with "wl" (or if two interfaces begin with "wlan" we will also have the same issue)? Should we make it specifiable from the user?

modules/sc-mesh-secure-deployment/src/server-mesh.py
```diff
@@ -40,7 +40,10 @@
     "mode": "mesh"  # "mesh=mesh network, ap=debug hotspot"
 }
 
-ADDRESSES = {'00:00:00:00:00:00': '10.20.15.1'}
+# ADDRESSES = {'00:00:00:00:00:00': '10.20.15.1'}
```
mbaddeley, line 43: Can remove this line.

modules/sc-mesh-secure-deployment/src/client-mesh.py
```diff
@@ -44,17 +47,45 @@ def get_data(cert_file, os):
...
     joined_list = ''.join(new_list)
     output_dict = joined_list.replace("\'", '"')
-    print('> Decrypted Message: ', output_dict)  # res =  json.loads(output_dict)
+
     return output_dict
 
 
+def verify_certificate(old, new):
+    """
+    Here we are validating the hash of the certificate. This is giving us the integrity of the file, not if the
```
mbaddeley, line 71: Should create a github issue for this

modules/sc-mesh-secure-deployment/src/server-mesh.py
```diff
@@ -87,12 +87,25 @@ def add_message(uuid):
         return 'Not Valid Certificate'
 
 
+def verify_certificate(old, new):
+    """
+    Here we are validating the hash of the certificate. This is giving us the integrity of the file, not if the
```
mbaddeley, line 95: Like the client-mesh.py - this should link to a github issue so we can PR it later.

modules/sc-mesh-secure-deployment/src/server-mesh.py
martin-tii: Right, I will leave it as 'wlan', so then we can think of another solution.

modules/sc-mesh-secure-deployment/src/server-mesh.py
martin-tii: done

mbaddeley (approved):

PR #27, opened 2021-05-31
Enable 11s for wpa_supplicant and wifi automatic detection for fogsw

joenpera:
11s:
- qca988x and ar9462 supported
- automatically detecting phy and device name
ibss:
- qca6174 and ar9462 supported
- automatically detecting phy and device name


pemanty (approved):

mbaddeley (approved):
nmcli removal may break something, but I will test and fix if this is the case.

PR #40, opened 2021-06-03
Feature/distributed

martin-tii:
This PR will not be merged, it's only for review purposes.

- Commit:

Fixing pandas crash, when installed by pip. Remove keyring locking (gui) when installing by pip

- Commit:

Update client on configure.sh to verify latency in case of multiple servers in the network.

- Commit:

------server-mesh.py-------
--> Added Subnet as initial parameter, default 10.0.0.0.
--> AuthServer flag added as a parameter in the Json ubuntu.
--> Currently, the 3rd node authenticated is added as server. This should be done in a smarter way.
--> Add_default_route was modified for the addition of the new subnet.
--> New store_authServer added. It creates a dataframe (AUTH_ROUTES) with the 'routing' info. It can be accessed as http://<ip_mesh>/authServ/

------README.md-------
--> Added a new diagram (distributed-auth.png) with an explanation about the distribution.

------client-mesh.py-------
--> Created a function (authServer) that will deal with the subnet IP (currently being random). It will start the 'configure.sh -s <subnet_IP>' and it will send this subnet IP to the server-mesh.py.
--> Code for config OpenWRT was removed
--> Verification of OS is only done for Ubuntu, otherwise, error message and exit.

------configure.sh------
--> Added a parameter in the server side to add subnet ip. In case the parameter does not appear use default addressing.


mbaddeley (commented):
Would it be better to use qperf to measure TCP rather than ICMP with ping? Or do we think ping is sufficient?

modules/sc-mesh-secure-deployment/src/server-mesh.py
```diff
@@ -65,9 +70,11 @@ def add_message(uuid):
         aux = aux_ubuntu if uuid == 'Ubuntu' else aux_openwrt
         if ip_mesh == IP_PREFIX + '.2':  # First node, then gateway
             aux['gateway'] = True
-            add_default_route(ip_address)  # we will need to add the default route to communicate
+            add_default_route(ip_mesh, ip_address)  # we will need to add the default route to communicate
         else:
             aux['gateway'] = False
+        if ip_mesh == IP_PREFIX + '.3':  # TODO: find smart way to set this value. Currently: only one server == '.3'
```
mbaddeley, line 76: A simple solution could be based on # of nodes joined who AREN'T servers. i.e. keep track of # of nodes joined then mod it so every 5th node (for example) is delegated as a server. 

Additionally, we need to think about scalability. There should probably be a max # of servers per mesh.

martin-tii:
no needed anymore

PR #57, opened 2021-06-20
User and mesh configuration is now taken from mesh_com.conf yaml file.

mbaddeley:
Moved the client and server mesh configuration settings to a configuration file (rather than being hardcoded in json in server-mesh.py)

- Added 'type' to configuration (11s or ibss). The corresponding script is now copied over on the client.
- Added a 'debug' configuration setting
- Added some client conf for set_hostname and disable_networking (useful for running a client on a dev machine if you don't want to mess up the hostname/networking)
- Modified the way that the client gets the unencrypted json from the server by concatenating the list and parsing for json using regex. Allows unlimited # of fields to be addded/sent from the server.

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/client-mesh.py
+++ b/modules/sc-mesh-secure-deployment/src/client-mesh.py
@@ -2,9 +2,11 @@
 
 import argparse
 import json
+import yaml
 import subprocess
 import time
 import hashlib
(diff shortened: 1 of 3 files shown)


martin-tii (approved):
LGTM

joenpera (approved):

PR #63, opened 2021-06-29
mesh_com:utils: Add phy parameter monitoring utility infra

govindsi:
Add utility infra to capture or periodically monitor/log various
mac/phy parameter that includes radio rssi, CSI, BW, RF temperature,
PER, peer stats etc.

This change adds RSSI monitoring support on a given interface with
configurable periodicity.

uses:monitor_phy_param.py -r period -i interface
log file: /var/logs/rssi2021-06-29 10:53:34.833729.txt

Signed-off-by: Govind Singh <govind.sk85@gmail.com>
Signed-off-by: Govind Singh <govind@ssrc.tii.ae>

### DIFF
--- a/modules/utils/monitor_phy_param.py
+++ b/modules/utils/monitor_phy_param.py
@@ -1,3 +1,4 @@
+import argparse
 import getopt
 import sys
 import datetime
@@ -31,22 +32,19 @@ def log_rssi():
         sleep(rssi_mon_interval)
 
 if __name__=='__main__':
-    argv = sys.argv[1:]
 
-    try:
-        opts, args = getopt.getopt(argv, "hr:i:")
-        for opt, arg in opts:
-            if opt in ['-h']:
-                print ("monitor_phy_param.py -r period -i interface")
-                sys.exit()
-            elif opt in ['-r']:
-                print ("rssi mon period:", arg)
-                rssi_mon_interval = int(arg)
-            elif opt in ['-i']:
-                print ("interface:", arg)
-                interface = arg
-    except getopt.error as err:
-        print (str(err))
+    # Construct the argument parser
+    phy_cfg = argparse.ArgumentParser()
+
+    # Add the arguments to the parser
+    phy_cfg.add_argument("-r", "--rssi_period", required=True, help="RSSI monitoring period Ex: 5 (equals to 5 sec)")
+    phy_cfg.add_argument("-i", "--interface", required=True)
+    args = phy_cfg.parse_args()
+
+    #populate args
+    rssi_mon_interval = int(args.rssi_period)
+    interface = args.interface
+
 
     Thread(target=log_rssi).start()
 
--- a/modules/utils/monitor_phy_param.py
+++ b/modules/utils/monitor_phy_param.py
@@ -13,6 +13,41 @@ from netaddr import *
 rssi_mon_interval = 5
 #default interface
 interface = "wlan0"
+mac_addr_filter = "FF:FF:FF:FF:FF:FF"
+
+def is_csi_supported():
+    soc_version_cmd = "cat /proc/cpuinfo | grep 'Revision' | awk '{print $3}'"
+    proc = subprocess.Popen(soc_version_cmd, stdout=subprocess.PIPE, shell=True)
+    soc_version = proc.communicate()[0].decode('utf-8').strip()
+    print("SOC VERSION:"+soc_version)
+    if ((soc_version == 'a020d3') or (soc_version == 'b03114')):
+        print("RPI nexmon CSI is supported, SOC VERSION:"+soc_version)
+        return  1
+    else:
+       return 0
+
+def capture_csi():
+    global interface
+    global mac_addr_filter
+    #Get CSI extractor filter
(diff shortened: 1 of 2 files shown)


modules/utils/monitor_phy_param.py
```diff
@@ -0,0 +1,52 @@
...
+import sys
+import datetime
+import subprocess
+from time import sleep
+from threading import Thread
+
+#defult rssi monitoring interval: 5sec
+rssi_mon_interval = 5
+#defualt interface
+interface = "wlan0"
```
martin-tii, line 11: For later, we need to think of a smart way to determine the interface. 
In some scripts I've used netifaces (https://pypi.org/project/netifaces/)
But also we need some logic to choose it

modules/utils/monitor_phy_param.py
```diff
@@ -0,0 +1,52 @@
...
+    rssi = proc.communicate()[0].decode('utf-8')
+    return rssi
+
+def log_rssi():
+    global rssi_mon_interval
+    fn_suffix=str(datetime.datetime.now())
+    log_file_path = '/var/log/'
+    log_file_name =  'rssi'+fn_suffix+'.txt'
+    while True:
+        f = open(log_file_path+log_file_name, 'a')
```
martin-tii, line 26: I prefer to use 'with' to avoid closing the file after (I always forget :P ), but it's only a matter of style.
`        with open(log_file_path + log_file_name, 'a') as f:
            rssi_sta = get_rssi()
            print(rssi_sta)
            f.write(rssi_sta)
        sleep(rssi_mon_interval)
        `

modules/utils/monitor_phy_param.py
```diff
@@ -0,0 +1,52 @@
+import getopt
+import sys
+import datetime
+import subprocess
+from time import sleep
+from threading import Thread
+
+#defult rssi monitoring interval: 5sec
+rssi_mon_interval = 5
+#defualt interface
```
martin-tii, line 10: typo in default

modules/utils/monitor_phy_param.py
```diff
@@ -0,0 +1,52 @@
+import getopt
+import sys
+import datetime
+import subprocess
+from time import sleep
+from threading import Thread
+
+#defult rssi monitoring interval: 5sec
+rssi_mon_interval = 5
```
martin-tii, line 9: Only asking, how should we determine the most optimal interval time?

modules/utils/monitor_phy_param.py
```diff
@@ -0,0 +1,52 @@
...
+        f.write(rssi_sta)
+        f.close()
+        sleep(rssi_mon_interval)
+
+if __name__=='__main__':
+    argv = sys.argv[1:]
+
+    try:
+        opts, args = getopt.getopt(argv, "hr:i:")
+        for opt, arg in opts:
```
martin-tii, line 38: isn't it better to use arg.parser? 
https://docs.python.org/3/library/argparse.html

martin-tii (approved):

modules/utils/monitor_phy_param.py
govindsi: This we need to derive based on few experiments with diff radio.
If we have motion awareness we can have aggressive interval in comparison to the static scenario.
This will be mostly overridden by the cmdline param.

modules/utils/monitor_phy_param.py
govindsi: will address in next patchset.

modules/utils/monitor_phy_param.py
govindsi: Yes, agreed.
As of now idea is to get this via args(-i IFNAME) as we can use this in various topology (mesh0/wlan0/p2p0/ibss0..etc..)

modules/utils/monitor_phy_param.py
govindsi: Yes, this seems easy to use, will change to argparser.

modules/utils/monitor_phy_param.py
mbaddeley: I put interface names into the conf file:

https://github.com/tiiuae/mesh_com/blob/develop/modules/sc-mesh-secure-deployment/src/mesh_com.conf

Can we pull from there?

govindsi: yeah, sure, it makes sense.
I will add new patchset to retrieve ifname from conf.

mbaddeley: Sorry, I actually meant from here. I completely forgot to push this.

https://github.com/tiiuae/mesh_com/blob/hotfix/fix_interface_list/modules/sc-mesh-secure-deployment/src/mesh_com.conf

govindsi: Created separate conf for phy param along with interface name.
As of now this module is just for experimental purpose(to study some ongoing research projects) , we can converge to main conf once we finalize on the scheme/radio/topology.

joenpera (approved):

govindsi:
Merging this series as this adds basic infra.
Incremental update is WIP.

PR #65, opened 2021-07-07
Setup fastrtps profile with mesh whitelisting address

jnippula:
ROS2 data distribution system discovery mechanism publishes topic
information by default to all network interfaces available in the
system. To avoid that the fastrtps profile needs to be updated to
define ip addresses for the interfaces allowed to be used for ROS2
topic communication.

### DIFF
--- a/modules/mesh_com/mesh_com/mesh_executor.py
+++ b/modules/mesh_com/mesh_com/mesh_executor.py
@@ -183,6 +183,10 @@ class MeshNetwork:
                              quote(self.settings.tx_power),
                              quote(self.settings.country)])
 
+            if os.path.exists("/opt/ros/foxy/setup_ros_profiles.sh"):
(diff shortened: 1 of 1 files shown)


jnippula:
There is related PR for fogsw_config providing the setup script:
https://github.com/tiiuae/fogsw_configs/pull/10

maseabunikie (approved):
Looks good.

jnippula:
Let's delay this change until we get more bandwidth to the mesh. With current low bandwidth mesh network we don't want to allow high ros2 topic message traffic to overload the mesh that would jam also the local ROS2 communication causing very high cpu load and loss of messages in critical topics.

joenpera:
We have delivery practice as follows: "own branch" -> develop -> main (no direct commits)

joenpera (changes requested):
This pull request needs to done to develop branch ( when decided to take in use ).

PR #67, opened 2021-07-15
mesh_parameters subscriber qos change

jnippula:
To verify that mesh_parameters are delivered reliably and the param message remains in the topic history even if the subscriber is create/established after the mesh_parameters message is published bu communication_link, the topic QoS settings needs to be defined to allow this behavior. As communication_link already publish to the topic with reliable way, the mesh_clom shall also subscribe the topic with same qos parameters to be able to read the message at all.

This PR changes the subscriber QoS to be align with Qos set by publisher in communication_link.

### DIFF
--- a/modules/mesh_com/mesh_com/mesh_subscriber.py
+++ b/modules/mesh_com/mesh_com/mesh_subscriber.py
@@ -1,7 +1,9 @@
 import rclpy
 from rclpy.node import Node
-from rclpy.qos import QoSPresetProfiles
-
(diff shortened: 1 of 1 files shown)


jnippula:
Functionality verified with drone setup on top of Rel4 content. Mesh parameters are properly received by mesh subscriber with defined qos settings

jlaitine (approved):

maseabunikie (approved):
Looks good.

PR #70, opened 2021-07-21
mesh_com: initial multi radio support script

joenpera:
- created new script for multi-radio purposes
- 1st radio follows json frequency parameter and currently second radio is hardcoded to 2412MHz(channel 1)
- Support any number of selected Qualcomm PCI wifi cards (frequency distribution needs to be solved)

example:
 sudo ./mesh-11s-mr.sh mesh 192.168.1.2 255.255.255.0 00:11:22:33:44:55 1234567890 mymesh2 5220 30 fi
 or
 sudo ./mesh-11s-mr.sh ap
 or
 sudo ./mesh-11s-mr.sh off

### DIFF
--- a/common/scripts/mesh-11s-mr.sh
+++ b/common/scripts/mesh-11s-mr.sh
@@ -19,11 +19,14 @@ Parameters:
   <phyname>    - optional
 
 example:
- sudo ./mesh-11s-mr.sh mesh 192.168.1.2 255.255.255.0 00:11:22:33:44:55 1234567890 mymesh2 5220 30 fi wlan1 phy1
- or
- sudo ./mesh-11s-mr.sh ap
- or
- sudo ./mesh-11s-mr.sh off
+ For One Radio:
+   sudo ./mesh-11s-mr.sh mesh 192.168.1.2 255.255.255.0 00:11:22:33:44:55 1234567890 mymesh2 5220 30 fi wlan1 phy1
+ Automatic detection:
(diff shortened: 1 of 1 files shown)


common/scripts/mesh-11s-mr.sh
```diff
@@ -295,39 +295,53 @@ main ()
...
+          phyname=$(echo "$pair" | cut -f1 -d ',')
+          wifidev=$(echo "$pair" | cut -f2 -d ',')
+
+          echo "  --------------------------------------------------------"
+
+          count=$((count+1))
+          echo "- #$count Found: $wifidev $phyname"
+
+          if [ "$count" -gt 1 ]; then
+            # ACS - Automatic Channel Selection, not yet working for wpa/ap
```
govindsi, line 322: Is ACS enable not enabled due to missing config parameter in supplicant conf or do u see issue with reported ACS survey data?

joenpera: I had acs=1 in the wpa_supplicant.conf, but this needs to be checked further why it isn't working with 11s.  It would be nice to set second radio to the best channel available.

mbaddeley (approved):
Approved. Can we create an issue for the hardcoded second radio freq though?

common/scripts/mesh-11s-mr.sh
govindsi: Thanks for the update, we can check if https://patchwork.kernel.org/project/linux-wireless/patch/1552044366-9186-1-git-send-email-pillair@codeaurora.org/ is present and track this via new issue.

govindsi (approved):

joenpera:
#72 created for the hard-coded second radio channel

PR #71, opened 2021-07-23
mesh_com: ROS node visualisation update

joenpera:
- topology delivered in json format
- batman mesh network information in json format
- better performance (topology and stats executed in separate thread)
- socket receiver buffer not anymore fixed size.  Every message has 4-byte prefix
  which includes message lenght. (msgs between mesh node and executor)


mbaddeley (approved):
Cool, nice PR!

pemanty (approved):

PR #74, opened 2021-07-29
Feature/ibss wpa2 rsn(RFC: DO_NOT_MERGE)

govindsi:
RFC:setup IBSS-RSN(WPA2) config on supported radio instead of WEP(highly vulnerable security mode).
This needs to be tested with more radio's to avoid compatibilities with invalid/fake capabilities from driver. Fake drivers can be patched to fix invalid reporting of the capabilities.

### DIFF
--- a/common/scripts/mesh-ibss.sh
+++ b/common/scripts/mesh-ibss.sh
@@ -90,7 +90,17 @@ echo "sudo mesh $1 $2 $3 $4 $5 $6 $7 $8 $9 ${10} ${11}"
         help
       fi
 
-cat <<EOF >/var/run/wpa_supplicant-adhoc.conf
+
+phy_if=$(iw dev $wifidev info | awk '/wiphy/ {printf "phy" $2}')
+rsn_ibss_cap=$(iw phy $phy_if info | awk '/RSN/ {printf $3}')
+
+if [ "$rsn_ibss_cap" = "RSN-IBSS." ]; then
+    ibss_conf_file="wpa_supplicant-adhoc-wpa2.conf"
+else
+    ibss_conf_file="wpa_supplicant-adhoc-wep.conf"
+fi
+
+cat <<EOF >/var/run/wpa_supplicant-adhoc-wep.conf
 ctrl_interface=DIR=/var/run/wpa_supplicant
 # use 'ap_scan=2' on all devices connected to the network
 # this is unnecessary if you only want the network to be created when no other networks..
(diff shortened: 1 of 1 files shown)


common/scripts/mesh-ibss.sh
```diff
@@ -108,6 +118,22 @@ network={
...
+ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev
+update_config=1
+ap_scan=1
+p2p_disabled=1
+country=$9
+network={
+    ssid="$6"
+    key_mgmt=WPA-PSK
+    proto=RSN
+    psk="1234567890"
```
joenpera, line 131: psk hardcoded and not taken from parameters.

common/scripts/mesh-ibss.sh
```diff
@@ -90,8 +90,18 @@ echo "sudo mesh $1 $2 $3 $4 $5 $6 $7 $8 $9 ${10} ${11}"
         help
       fi
 
-cat <<EOF >/var/run/wpa_supplicant-adhoc.conf
-ctrl_interface=DIR=/var/run/wpa_supplicant
+
+phy_if=$(iw dev $wifidev info | awk '/wiphy/ {printf "phy" $2}')
+rsn_ibss_cap=$(iw phy $phy_if info | awk '/RSN/ {printf $3}')
+
+if [ "$rsn_ibss_cap" = "RSN-IBSS." ]; then
```
joenpera, line 97: Even qca6174 reports RSN-IBSS support, not working.  Do not merge before this is solved!

joenpera (changes requested):

common/scripts/mesh-ibss.sh
govindsi: Yeah, we need to fix ath10k bug.
If its FW bug, it will be long pole to fix.
Right now iwl and ralink radio is tested.

govindsi: I am able to repro with QCA6174<------>ralink and RM4.4.XXX FW is drooping the EAPOL frame(M1).
Same observation is seen with the logs shared by you.
I have initiated one thread in ath10k upstream group, we can revisit this once we have fix available for QCA6174 or we can explicitly enable to known working radios(in future).

http://lists.infradead.org/pipermail/ath10k/2021-August/012738.html

joenpera:
Should we close this pull request without merge?   No activity in the ath10k upstream e-mail group.

govindsi:
> Should we close this pull request without merge? No activity in the ath10k upstream e-mail group.

Yeah, as we don't plan to use IBSS anymore. Hence, we can close.

PR #75, opened 2021-08-02
Revert "mesh_com: Store certificate in system keystore"

govindsi:
Reverts commit 603597872f63f2c49b8ebbeded0da0e2144738c3 as
it needs root user in server side. Explore different mechanism
using selinux policies with user group to avoid local provisioning
of the key.

Signed-off-by: Govind Singh <govind.sk85@gmail.com>


mbaddeley (approved):

modules/sc-mesh-secure-deployment/configure.sh
```diff
@@ -234,10 +234,8 @@ function server {
   popd
   # Advertise the server using avahi (zeroconf)
   avahi-publish-service mesh_server _http._tcp 5000 &
-  #Install certificate
```
martin-tii, line 237: we should do chmod to non-root user before copy to /etc

martin-tii (approved):

modules/sc-mesh-secure-deployment/configure.sh
govindsi: Yup, this will work as WAR. Will explore if there is better way to handle the same.
We can try with selinux policies in hardened OS (if selinux enabled).

govindsi:
revert and explore if there is better way to handle the same.
We can try with selinux policies in hardened OS (if selinux enabled).

PR #79, opened 2021-08-04
mesh_com: mtu size fixed for 11s

joenpera:


joenpera:
without this mtu size fix larger ip packages not delivered through.  e.g. iperf tests failing.

mbaddeley (approved):

liljaj (approved):

PR #87, opened 2021-09-14
Feature/docker

martin-tii:
Initial docker (ubuntu) running under Secure OS and Ubuntu OS.

- All the services have been modified to 'init.d' style.

- No need to reboot the node. 

- On client-mesh.py, python 'bash' scripts were removed and added to a new folder (src/bash). Now on the client-mesh.py only are executed these bash scripts.


joenpera (changes requested):
mesh-ibss.sh changes will break current fog drone ROS functionality.  If these are merged forwards Mesh network will stop working in drones.  Be aware that new release is built automatically when pushed to main.

martin-tii (commented):
@joenpera, these 2 lines are failing in Secure OS (path is not present).
What is your suggestion?
Create a new 'mesh-ibss2.sh' without this or find a way to make it work on Secure OS?

joenpera:
> @joenpera, these 2 lines are failing in Secure OS (path is not present).
> What is your suggestion?
> Create a new 'mesh-ibss2.sh' without this or find a way to make it work on Secure OS?

I can see that you always call the mesh-ibss.sh script with 10th parameter <interface>, it will skip the auto detection which supports only selected pci radios and that seems to be the behavior which is wanted in this case.  I can't see problem if you remove those changed lines.

I tried with SecureOS and with original version without any feature/docker - changes:
./mesh-ibss.sh mesh 192.168.1.2 255.255.255.0 00:11:22:33:44:55 1234567890 mymesh 5220 30 fi wlan1 phy1

Works fine.

joenpera (approved):
mesh-ibss.sh looks ok now.

mbaddeley (approved):

PR #98, opened 2021-10-18
Feature/docker

martin-tii:
Commit with all dependencies installed while docker execution:
* Better verification of docker environment (line 244) configure.sh
* clang libssl-dev were added to apt install on Dockerfile
* client and server requirements are installed during docker creation
* cryptolib is compiled during docker creation

During make client/server is verified if it's being executed on a docker environment. If negative, it installs all the requirements, otherwise, only cp crypto binaries.


mbaddeley (dismissed):

govindsi (dismissed):

martin-tii:
I've added all the following commit on develop, it shouldn't conflict anymore.
Can you please review it again, please? @mbaddeley @govindsi

govindsi (dismissed):

mbaddeley (approved):

govindsi (approved):

joenpera:
mesh_com build failed with this commit: 

https://github.com/tiiuae/mesh_com/runs/4421323222?check_suite_focus=true

PR #100, opened 2021-10-18
Feature/drone remote

govindsi:
Drone remote ID Changes(L0)

Supported:
MESH + AP DRI bcaster/observer support.
role selection from config(observer/broadcaster)
DRI data from as static file is appended in beacon vendor IE.
Basic BLE broadcaster(hardcoded UUID)

Not supported/Planned/TBD(L2):
ROS topic subscription to get DRI data.
data format conversion(ASTM format)
Security aspects.

### DIFF
--- a/modules/utils/python/remote_droneID.py
+++ b/modules/utils/python/remote_droneID.py
@@ -3,107 +3,108 @@ from time import sleep
 from threading import Thread
 import yaml
 
-#ROS node to convert to hex format
-#<element id><len><oui><vendor ie>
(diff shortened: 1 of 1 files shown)


joenpera:
Observer mode needs some code changes when taken in use with ROS.  dri_oui_list needs to be stored to internal variable for reprocessing purposes and get-method for the data.  Now observer is only printing scan results to threads stdout.   

Some requirements comes from ROS integration:
- ROS node is executed as user
- DRI needs to be executed as root

These can be implemented in a later phase when taken in use.

PR #101, opened 2021-10-19
Mesh ibss  and init.d script autostart fix

govindsi:


modules/sc-mesh-secure-deployment/conf/ap.conf
```diff
@@ -1,7 +1,4 @@
-  network={
-    ssid="AP-TEST"
-    mode=2
-    key_mgmt=WPA-PSK
-    psk="ssrcpassword"
-    frequency=2437
+network={
+  ssid="Anshul"
```
govindsi, line 2: My bad, will restores this file to original

govindsi: done

PR #115, opened 2022-01-19
fix container building

jnippula:


jnippula:
Move to develop branch

PR #116, opened 2022-01-20
Fix containerization of mesh-com

jnippula:
Dockerfile and github workflow created to make container to run in HardenedOS env.
There are one docker image created tii-mesh-com that can be started for running the mesh_executor or the pub&sub nodes.

```
docker run tii-mesh-com         --> start pub&sub
docker run tii-mesh-com init    --> start mesh executor

```


.github/workflows/tii-mesh-com.yaml
```diff
@@ -0,0 +1,47 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build tii-mesh-com image and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./modules/mesh_com/Dockerfile
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
```
joonas-fi, line 47: please also add:

```
labels: ${{ steps.meta.outputs.labels }}
```

Labels carry OCI-compliant container metadata, along with them is Git revision. The Git revision is required for `$ fog manifest update-component` tool to update both the container image reference and Git revision reference at the same time.

modules/mesh_com/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN apt-get update -y && apt-get install -y --no-install-recommends \
+    wpasupplicant \
+    wifi-firmware \
+    iw \
+    batctl \
+    alfred \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV PACKAGE_NAME mesh_com
+CMD [""]
```
joonas-fi, line 37: If this is required, it would be good to add in the Dockerfile a comment why, so the next person understands why this is here

modules/mesh_com/package.sh
```diff
@@ -5,6 +5,8 @@ if [ "${ROS_DISTRO}" = "" ]; then
    exit 1
 fi
 
+export DEB_BUILD_OPTIONS=nocheck
```
joonas-fi, line 8: This could use a code comment explaining why

.github/workflows/tii-mesh-com.yaml
jnippula: Added labels

modules/mesh_com/entrypoint.sh
```diff
@@ -1,9 +1,22 @@
 #!/bin/bash
```
joonas-fi, line 1: Should this be ```bash
#!/bin/bash -eu
```

It has the same effect as `set -eu`

Where the `e` means that the script will stop if any command fails. It is probably not intended behaviour that if `$ mesh-ibss.sh ap` fails we still continue to starting `mesh_executor`

modules/mesh_com/Dockerfile
jnippula: CMD removed as it is not really needed here

modules/mesh_com/package.sh
jnippula: Removed the nocheck build option. Build works without it. I must have added some specific changes when it failed, but now it seems to work fine anyway, so no extra option needed.

modules/mesh_com/entrypoint.sh
jnippula: You are right. That is needed to make failing visible

jnippula: Left unbound variable check out

joonas-fi (approved):
Looks good, nice work!

maseabunikie (approved):

joenpera (approved):
Agreed with @jnippula that we will merge this forward for test purposes and fixes will follow in coming release.

PR #118, opened 2022-01-28
Containerization fixes

jnippula:
Container image creation fixed to properly run mesh_com nodes in docker container with ROS2 galactic environment.


common/scripts/mesh-ibss.sh
```diff
@@ -126,6 +126,8 @@ EOF
       echo "$wifidev down.."
       iw dev "$wifidev" del
       iw phy "$phyname" interface add "$wifidev" type ibss
+      # same MTU as for $wifidev. without this the kernel warned that 1500 is too small to handle batman
+      ifconfig "$phyname" mtu 1560
```
joonas-fi, line 130: Please carefully review this

joenpera: We are not using mesh-ibss.sh for Mesh anymore.  mesh_com API version 2 should be used which mean 11s based Mesh network --> mesh-11s.sh is used in that case.

joonas-fi: This is in line with the software that runs on FOG Ubuntu: https://github.com/tiiuae/fogsw_systemd/blob/9a5bed31547b46ddf90139763ec8099585d0d759/system/mesh-init.service#L10

We are trying to replicate that in a Docker container.

I appreciate that newest version of mesh_com may have moved on, and we need to take that account in the future, but did you comment this meaning that this is still wrong, even in "older release version of mesh" context? This script is what FOG Ubuntu has been tested against, and AFAIK it has been tested as working.

If the script is what should work, I propose we use it for now and when we get a working container baseline, we'll do separate work to upgrade to your latest approach? We can of course defer merging this branch until it works against your newest approach.

joenpera: To simplify:
- your changes are in mesh-ibss.sh Mesh configuration section which is not used
- mesh-ibss.sh is used only for debug AP setup in device startup before Mesh setup
- currently mesh setup is done by using API v2 and it will use mesh-11s.sh (this has been in mesh_com releases already for months).  Of course correct parameters are needed from cloud side to use API v2.

joenpera (approved):

pemanty (dismissed):

joenpera (changes requested):
Just realized that this pull request is to main.  Pull needs to be done to develop.

.github/workflows/tii-mesh-com.yaml
```diff
@@ -43,6 +41,6 @@ jobs:
         with:
           context: .
           file: ./modules/mesh_com/Dockerfile
-          push: ${{ github.event_name != 'pull_request' }}
+          push: true
```
jnippula, line 44: Should this change be reverted to avoid unnecessary big amount of images pushed into ghcr during some feature testing/developing/reviewing?

jnippula: Maybe we go with this change now. Need more discussion with bigger audience what is the policy regarding image push to ghcr

joenpera (approved):

pemanty (approved):

PR #132, opened 2022-03-07
Feature/sec os 11s

govindsi:


martin-tii (approved):

common/scripts/mesh-11s.sh
```diff
@@ -91,6 +91,7 @@ echo "sudo mesh $1 $2 $3 $4 $5 $6 $7 $8 $9 ${10} ${11}"
         help
       fi
 
+iw dev "$wifidev" set mesh_param mesh_fwding 1
```
joenpera, line 94: Should be:
iw dev "$wifidev" set mesh_param mesh_fwding 0

joenpera (changes requested):

common/scripts/mesh-11s.sh
joenpera: fixed with [55e2be6](https://github.com/tiiuae/mesh_com/pull/132/commits/55e2be66d7497aa10eca90288344f5ef7da7db9e)

joenpera (approved):

PR #133, opened 2022-03-08
Feature/develop pgp

martin-tii:
Dynamic GW selection added as a function.
gw folder with main and tools added


martin-tii:
Also, here we have some code for mesh 1.5. If you think it's not necessary to merge, I can do a unique PR only for the dynamic GW.

mbaddeley (approved):

govindsi (approved):

PR #138, opened 2022-03-20
Probe Request flood attack scripts

aravindraja93:
- Have created a new directory called "python_scripts" to store the python files.
- The attack bash script calls the python script which uses scapy to flood channel with PR request


govindsi (approved):

common/tests/security/python_scripts/flood_pr.py
```diff
@@ -0,0 +1,49 @@
...
+# Combining the RadioTap and the DOT11 frames, the probe request packet is crafted
+# User input is used to stop the flooding currently
+
+pr_pkt = radiotap/dot11_frame
+
+try: 
+    while True:
+        sendp(pr_pkt,iface=interface)
+except KeyboardInterrupt:
+    print("Stopped transmetting fake ProbeRequest frames")
```
martin-tii, line 48: typo

common/tests/security/pr_floodattack.sh
```diff
@@ -0,0 +1,98 @@
...
+
+	_init_result
+	
+	if [ "$result" -eq "1" ]; then
+	
+		echo "Device already in monitor mode"
+	
+	else
+		airmon-ng check kill
+		airmon-ng start $interface
```
martin-tii, line 31: should we use airmon-ng here? 
or leave it more generic with `iwconfig`
@govindsi ??

martin-tii (commented):
Only to check if we should go with` airmon-ng` or with `iwconfig`

common/tests/security/pr_floodattack.sh
aravindraja93: Have changed to use the `iwconfig`

common/tests/security/python_scripts/flood_pr.py
aravindraja93: Fixed this and other typos

martin-tii (approved):

PR #190, opened 2022-09-02
Battery temp to field test logger

jihonen:
Adding battery temperature measurement to field test logger.


saauvine (commented):
Implementation should be probably enhanced a bit to be backwards compatible with log files that do not contain new battery temp column.

joenpera (commented):
Logger part is ok.  Plotter could have changes Saku mentioned.

saauvine (approved):

joenpera (approved):

PR #192, opened 2022-09-06
Ignore bad lines in CSV log file

saauvine:
Pandas dataframe is created using option to ignore (but warn)
bad lines in csv file. Possible incomplete log lines are dropped
from dataframe before actual processing.

Jira-Id: WC-148

Signed-off-by: Saku Auvinen <saku.auvinen@unikie.com>


common/tools/field_test_log_plotter/ftl_plotter.py
```diff
@@ -94,6 +94,9 @@ def __create_dataframe(self):
                                                         inplace=True)
         # Drop any DF lines where timestamp is NaN
         self.df.dropna(subset=['Timestamp'], inplace=True)
+        # Drop corrupted DF lines i.e. the ones where the
+        # last expected csv file column is empty.
+        self.df.dropna(subset=['3v3 current [mA]'], inplace=True)
```
joenpera, line 99: How about using "df.columns[-1]" for last column selection?  Instead of hardcoded column name.

saauvine: That is definitely better way to do it.

joenpera (approved):

jihonen (approved):

PR #193, opened 2022-09-14
Feature/1.5

anshul-tii:
A script to enable connectivity with a desired AP


common/scripts/mesh-11s.sh
```diff
@@ -107,7 +107,7 @@ network={
     psk="$5"
     key_mgmt=SAE
     ieee80211w=2
-    mesh_fwding=0
```
joenpera, line 110: "mesh_fwding=0" configuration can be used with our wpa_supplicant.  If this is commented out, it breaks other projects which are using common scripts.

joenpera (changes requested):
Just checking the common scripts part.

common/scripts/mesh-11s.sh
joenpera: af3cfe5 fixes this one.

anshul-tii:
Still not integrated on develop

PR #205, opened 2022-12-05
Support qos-olsr and enable logging

md3w:
- Add option for qos-olsr routing protocol
- Pipe output of qos-olsr and olsr to a log file in /tmp


joenpera (changes requested):
PRs to develop branch only.

PR #227, opened 2023-02-14
Bridge certain ethernet port with mesh network based on detected HW

saauvine:
Jira-ID: WC-210 mesh_com_1.0: Ethernet eth0 not bridged with mesh network

Detected HW version is brought visible inside docker by copying /etc/comms_pcb_version file into container-data. HW version is used to select what ethernet port is bridged with mesh network.

### DIFF
--- a/modules/utils/docker/entrypoint.sh
+++ b/modules/utils/docker/entrypoint.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 
-COMMS_PCB_VERSION_FILE="/opt/comms_pcb_version"
+COMMS_PCB_VERSION_FILE="/opt/hardware/comms_pcb_version"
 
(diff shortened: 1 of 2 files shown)


martin-tii (approved):

joenpera (approved):

govindsi (approved):

modules/utils/docker/entrypoint.sh
```diff
@@ -1,5 +1,7 @@
 #! /bin/bash
 
+COMMS_PCB_VERSION_FILE="/opt/hardware/comms_pcb_version"
```
govindsi, line 3: This is just for future tracking, we can combine all required config for the container later with different subnodes

PR #239, opened 2023-03-06
spectral capture compatible with ath9k ath10k

dania-tii:


modules/sc-mesh-secure-deployment/src/1_5/common/test/spectral_capture.py
```diff
@@ -1,14 +1,35 @@
...
+    
+    vals_dict = []
+    vals_dict = spec.get_values()
+
+    vals_list = []
+    for key, value in vals_dict.items():
+       vals_list.append(list(value))
+
+
+    driver = os.popen('ls /sys/kernel/debug/ieee80211/phy* | grep ath').read().strip()
```
govindsi, line 25: Move this to read function, read needs to return pd frame, later to be used by ML class

PR #251, opened 2023-03-14
DO_NOT_MERGE:TEST:USSP_ASTM_RID

govindsi:


modules/mesh_com/Dockerfile
```diff
@@ -47,8 +47,10 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends \
     rfkill \
     iproute2 \
     python3-pip \
+    vim \
```
govindsi, line 50: This is only for debugging

aravindraja93: Yes, Wanted to keep it till we verify the working and then remove in the final step

PR #262, opened 2023-03-23
Hotfix for issue with MCC startup in MS v1.0

mbaddeley:
Adding the sleep allows the AP to come up properly. Only needed for v1.0

### DIFF
--- a/modules/utils/docker/entrypoint.sh
+++ b/modules/utils/docker/entrypoint.sh
@@ -77,6 +77,99 @@ psk="ssrcdemo"
 EOF
 }
 
+#write in the dhcpd.conf and the olsrd.conf add the ipv4 subnet and the random ipv6
+create_dhcpd_config()
+{
+  SUBNET="$1"
+
+  cat > /etc/dhcp/dhcpd.conf <<- EOF
+    default-lease-time 600;
+    max-lease-time 7200;
+    ddns-update-style none;
+    authoritative;
+
+    subnet $SUBNET.0 netmask 255.255.255.0 {
+            range $SUBNET.100 $SUBNET.199;
+            option routers $SUBNET.1;
+    }
+  EOF
+  cp /dev/null /var/lib/dhcp/dhcpd.leases
+}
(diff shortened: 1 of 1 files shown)


joenpera (approved):
Approved even I don't like the sleep() usage.

modules/utils/docker/entrypoint.sh
```diff
@@ -98,7 +192,20 @@ elif [ "$mode" = "ap+mesh_mcc" ]; then
   # Set frequency band and channel from given frequency
   calculate_wifi_channel "$ch"
   ifconfig $ifname_ap up
+  
+  #random generate the ipv6
+  IPV6_PREFIX=$( echo fd`dd if=/dev/urandom bs=7 count=1 status=none | xxd -p` | sed 's/\(....\)/\1:/g' )
+  ip addr add $SUBNET.1/24 dev br-lan && ip addr add $IPV6_PREFIX:1/64 dev br-lan
+
+  create_olsrd_config "wlp1s0" "$SUBNET" "$IPV6_PREFIX"
```
govindsi, line 200: We can move  olsr config update under  "$algo" = "olsr" , may be this can go in next PR.

modules/utils/docker/entrypoint.sh
```diff
@@ -98,7 +192,20 @@ elif [ "$mode" = "ap+mesh_mcc" ]; then
...
+  # FIXME: launch olsrd
```
govindsi, line 201: not needed as its getting launched below
 "$algo" = "olsr" 
(olsrd -i wlp1s0 -d 0)&
....

govindsi (approved):

mbaddeley:
For some reason all the QoS-OLSR changes are in this, I must have branched off the wrong thing. I changed one line, not 100! Will fix.

mbaddeley:
Fixed so it no longer has all the other crap in it. It's now just the sleep.

govindsi (approved):

PR #267, opened 2023-04-03
detect mesh interface channels available to scan

dania-tii:


modules/sc-mesh-secure-deployment/src/1_5/common/test/config_spectralscan.py
```diff
@@ -3,60 +3,84 @@
...
+        None
 
+# Temporary workaround for mesh scanning: mesh interface cannot scan it's own channel so remove it temporarily
+all_bands = ' '.join(str(ch) for ch in channel_list)
 all_bands = all_bands.replace(channel_freq, '').strip()
-all_2_4 = all_2_4.replace(channel_freq, '').strip()
-all_5 = all_5.replace(channel_freq, '').strip()
 
+# Get 2_4GHz and 5GHz channels from full list of channels retreived
+all_5 = [5180, 5200, 5220, 5240, 5260, 5280, 5300, 5320, 5745, 5765, 5785, 5805, 5825, 5845] #check if any one of 5ghz channels are in available list of channels to scan
```
govindsi, line 60: iw list will give u the list of allowed channel based on regulatory and CC settings, you can parse and populated from there.

dania-tii: yes that's being done in lines 37 - 55! It will return a full list of 2.4 and 5 GHz channels that are available to scan, Returned as "channel_list"

The "all_5" list you referenced is just looped through to find in the "channel_list" the first instance of a 5 GHz channel in order to split the full list of 2.4 and 5 into:
- 2.4 band channels
- 5 band channels
separately in the case of a band specific scan

govindsi (approved):

modules/sc-mesh-secure-deployment/src/1_5/common/test/config_spectralscan.py
dania-tii: I've updated the code in order to split into 2.4 and 5 bands dynamically

PR #268, opened 2023-04-03
Mesh com 1.5 compatibility

SelinaTII:
** mesh.conf --> Added BRIDGE=true to enable bridging mode (put false to disable bridge)
** mesh_com_11s.conf --> Added bridge, meshint to denote bridging mode and mesh interface
** entrypoint.sh --> Added condition: if BRIDGE = true in mesh.conf, write bridge: True, meshint: br-lan to mesh_com_11s.conf, else write bridge: False, meshint: bat0
** ConnectionMgr.py: Added attribute bridge to class ConnectionMgr which is read from mesh_com_11s.conf. In starting_mesh(), run bridge_settings() after starting mesh if bridge is true
** mesh_utils.py --> Imported yaml, added function get_mesh_int() to get the interface from meshint in mesh_com_11s.conf, assigned the value returned by get_mesh_int() as default mesh interface in arguments to other functions
** header.py --> Replaced hardcoded value of MESHINT with mesh_utils.get_mesh_int()
** continuous/utils.py --> Replaced hardcoded value of MESHINT with mesh_utils.get_mesh_int()
** continuous/ca_main.py --> Replaced hardcoded value of MESHINT with mesh_utils.get_mesh_int()
** quarantine.py --> Replaced hardcoded interface 'bat0' with mesh_utils.get_mesh_int()
** mutual.py --> Updated my_ip_mesh to be the br-lan ip after the bridge is established

Jira_Id : MSS20-125

Signed-off-by: Selina Shrestha <selina.shrestha@tii.ae>


martin-tii (approved):

modules/utils/docker/mesh.conf
govindsi, line 1: config is sourced from /opt/mesh.conf in entrypoint.
source /opt/mesh.conf

We will have to use single config for both 1.0/1.5 version. Probably we can use new config for all modes.

martin-tii (changes requested):

modules/utils/docker/entrypoint.sh
```diff
@@ -30,6 +28,7 @@ configure()
       mesh_if=$MESH_VIF
       meshVersion=$MSVERSION
       ML=$ML
+      bridge=$BRIDGE
```
SelinaTII, line 31: @govindsi  @martin-tii  I've changed it to get bridge mode from /opt/mesh.conf now

Monika-Research (approved):
Tested eth mode and it works fine. MCC mode needs to be enabled as well and entrypoint.sh needs to re-structured accordingly. However, it can be a new request.

martin-tii:
DO NOT MERGE, 
This PR will be closed. 
Use https://github.com/tiiuae/mesh_com/pull/273 instead

martin-tii:
Replaced with [PR 273](https://github.com/tiiuae/mesh_com/pull/273 )

Issue #270, opened 2023-04-06
github actions are failing due to artifactory location change


govindsi:
#9 7.455 E: Failed to fetch https://ssrc.jfrog.io/artifactory/ssrc-debian-public-remote/dists/focal/fog-sw/binary-amd64/Packages  404   [IP: 20.101.176.184 443]
#9 7.456 E: Some index files failed to download. They have been ignored, or old ones used instead.

ERROR: failed to solve: process "/bin/bash -c apt-get update -y && apt-get install -y --no-install-recommends     wpasupplicant=2.9.0-20-6~git20210701.43d8971     wifi-firmware     iw     batctl     alfred     rfkill     iproute2     && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 100
Error: buildx failed with: ERROR: failed to solve: process "/bin/bash -c apt-get update -y && apt-get install -y --no-install-recommends     wpasupplicant=2.9.0-20-6~git20210701.43d8971     wifi-firmware     iw     batctl     alfred     rfkill     iproute2     && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 100
/***********************************/

maseabunikie:
There are fixes available for this issue in the branch:
https://github.com/tiiuae/mesh_com/blob/feat-doodle/modules/mesh_com/Dockerfile

In addition this line needs to be updated:
https://github.com/tiiuae/mesh_com/blob/feat-doodle/modules/mesh_com/Dockerfile.build_env#L30
Use the URL: https://ssrc.jfrog.io/artifactory/ssrc-deb-public-local

maseabunikie:
Commits to be added are the following.
From: https://github.com/tiiuae/mesh_com/commit/b459d65a78359a43763f5e43e77b8c47e5ed8854
To: https://github.com/tiiuae/mesh_com/commit/73b3a424173812a758d74deb526addc95c555faa

Please, check the commits carefully because they might have conflicts.

aravindraja93:
The Docker build succeeds, 

The deb build for Humble release is failing.

Tried updating the dists reference from focal->jammy locally,  in the modules/mesh_com/pacakge.sh

```
I: pybuild base:239: cd /mesh_com/sources/modules/mesh_com/.pybuild/cpython3_3.10/build; python3.10 -m pytest 
[...]

PR #271, opened 2023-04-06
CherryPicked commits to enable upgrade to latest base image

aravindraja93:
PR to solve this issue : https://github.com/tiiuae/mesh_com/issues/270

### DIFF
--- a/modules/mesh_com/entrypoint.sh
+++ b/modules/mesh_com/entrypoint.sh
@@ -1,16 +1,56 @@
 #!/bin/bash -e
 
-# . /enclave/drone_device_id
-
-# export DRONE_DEVICE_ID
-
-source /opt/ros/galactic/setup.bash
+source /opt/ros/humble/setup.bash
 
 # I don't know what we're doing wrong, but Python isn't able to resolve mesh packages without this.
 # (other Python packages seem to reside under /usr/lib/python3/dist-packages)
-export PYTHONPATH=/opt/ros/galactic/lib/python3.8/site-packages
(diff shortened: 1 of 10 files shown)


modules/mesh_com/entrypoint.sh
maseabunikie, line 1: @joenpera @govindsi 
Please, take a look to this file carefully. This is taken from mesh chain changes done last year. It is possible that you don't want this on the main branch.

joenpera: This needs to be asked from @sergei-petroff.  He knows the setup which should be supported in mainline.

joenpera (approved):

govindsi (approved):

govindsi (commented):
Seems workflow is still failing.

 cp: cannot stat 'modules/*.deb': No such file or directory
Error: Process completed with exit code 1.

govindsi (changes requested):
Seems its still require some more changes to fix the workflow.

 cp: cannot stat 'modules/*.deb': No such file or directory
Error: Process completed with exit code 1.

aravindraja93:
@govindsi : Debian build script updated. 

And also the deb workflow is repeating having alternate scripts for the same goal , 

These two different scripts are used for the same goal, Better to fix using the same one:

1. https://github.com/tiiuae/mesh_com/blob/main/modules/mesh_com/package.sh
2. https://github.com/tiiuae/fog-ros-baseimage/blob/main/builder/packaging/package.sh

govindsi (approved):
Looks good now.

PR #276, opened 2023-04-24
Fix container double start/reset (part2)

joenpera:
accidentally left original code outside the if-statement.

Jira-Id: WC-225


Monika-Research (approved):
Works fine now. This fixes the MCC AP issue as well.

Monika-Research (approved):

saauvine (approved):

PR #278, opened 2023-04-28
MCC, bridge mode compatibility with MS 1.5 and improvisations from demo

SelinaTII:


joenpera (approved):
Just general note that we have work in progress with entrypoint.sh and it will cleaned out.   In a future it will include just process starts.  This means that all network related configurations are in one script (mesh-11s.sh) and it takes configuration file as parameter.   Work in ongoing in feature/nats-integration - branch.

govindsi (approved):

Monika-Research (approved):

PR #289, opened 2023-05-30
Fix e2e connectivity between laptop->csl->csl->laptop.

mbaddeley:
Set bat0 and br-lan MTUs to 1500 rather than 1460. This fixes end to end connectivity between connected devices (e.g., laptops/phones) on the AP or eth bridge (using 1460 only allows laptop->csl->csl.....).

### DIFF
--- a/common/scripts/mesh-11s.sh
+++ b/common/scripts/mesh-11s.sh
@@ -164,7 +164,7 @@ EOF
         echo "bat0 ip address.."
         ifconfig bat0 "$2" netmask "$3"
         echo "bat0 mtu size"
-        ifconfig bat0 mtu 1460
+        ifconfig bat0 mtu 1500
         echo
         ifconfig bat0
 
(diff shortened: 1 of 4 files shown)


common/scripts/mesh-11s.sh
```diff
@@ -164,7 +164,7 @@ EOF
         echo "bat0 ip address.."
         ifconfig bat0 "$2" netmask "$3"
         echo "bat0 mtu size"
-        ifconfig bat0 mtu 1460
+        ifconfig bat0 mtu 1500
```
govindsi, line 167: may cause regression in drone side, probably overriding in entrypoint.sh will be good idea.

mbaddeley: Reverted and made changes to the bridge and mcc settings scripts instead.

govindsi (approved):

joenpera (approved):

PR #302, opened 2023-09-05
QoS OLSR v2

francislin12:
To summarize our changes : 
1- we assigned a static IP address on the wlp1s0 interface to 192.168.11.  for the QoS-OLSR 
2- we  assigned a SUB NET  for QoS-OLSR or the br-lan to allow routing with the AP
3- we allowed QoS-OLSR to route on IPV4 or IPV6
4- we allowed QoS_OLSR to run on the background and to write into the configuration file OLSRD.conf

### DIFF
--- a/modules/utils/docker/entrypoint.sh
+++ b/modules/utils/docker/entrypoint.sh
@@ -152,20 +152,11 @@ install_packages()
 mesh_service()
 {
   #start mesh service if mesh provisioning is done
-  hw_platform=$(grep Model /proc/cpuinfo| awk '{print $5}')
-  if [ "$hw_platform" == "Compute" ]; then
-    if [ -f "/opt/S9011sMesh" ]; then
-      #start Mesh service
-      echo "starting 11s mesh service"
-      /opt/S9011sMesh start
-      sleep 2
-    fi
-  else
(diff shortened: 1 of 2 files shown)


joenpera (changes requested):

mbaddeley:
@joenpera @govindsi I believe we can revert the changes to entrypoint.sh as these only are for SCC mode and entrypoint.sh will change anyway? Either way, we also don't use SCC mode at the moment.

joenpera:
> @joenpera @govindsi I believe we can revert the changes to entrypoint.sh as these only are for SCC mode and entrypoint.sh will change anyway? Either way, we also don't use SCC mode at the moment.

New entrypoint_nats.sh (new default entrypoint) includes only service startups and network configurations are done in started services.  Improves readability a lot.

mbaddeley:
Ok latest commit has been tested end to end with two sleeves and two phones on BOTH qos-olsr AND batman-adv. I'm happy to recommend a merge now if @joenpera and @govindsi are happy with everything else?

mbaddeley (approved):
Approving as has been tested e2e with both BATMAN and QoS-OLSR in MCC mode.

joenpera (approved):

PR #303, opened 2023-09-06
Dp 6746 mesh com entrypoint refactoring

mtpaasto:


modules/mesh_com/entrypoint.sh
```diff
@@ -94,7 +86,5 @@ else
 
     # Start mesh publisher
     ros2 run mesh_com mesh_publisher --ros-args -r __ns:=/$DRONE_DEVICE_ID &
```
maseabunikie, line 88: Remove & in the end of the command to leave the ros node running and keep container alive.

mtpaasto: This is being addressed in another PR: https://github.com/tiiuae/mesh_com/pull/304/
Perhaps I should combine these both in to one?

mtpaasto:
Work moved to https://github.com/tiiuae/mesh_com/pull/309

PR #304, opened 2023-09-06
DP 6747 enable mesh publisher

mtpaasto:


mtpaasto:
Work moved to https://github.com/tiiuae/mesh_com/pull/309

PR #305, opened 2023-09-06
Nfs mesh coms automation

govindsi:
Allow default mesh service with network boot flavour

### DIFF
--- a/modules/utils/docker/entrypoint.sh
+++ b/modules/utils/docker/entrypoint.sh
@@ -435,6 +435,7 @@ EOF
     /bin/bash "$BRIDGE_SETTINGS"
     fi
 fi
+fi
 
 #start comms sleeve web server for companion phone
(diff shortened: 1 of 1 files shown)


DaniilTroshkovTII (approved):

modules/utils/docker/entrypoint.sh
```diff
@@ -416,7 +425,15 @@ EOF
...
     mesh_service
+    # Start only mesh service with out bridge for network boot as ethernet port will be dedicated
+    # to network boot.
+    if [ -f /root/build.info ]; then
+            if grep -Fxq "TCDIST_PLATFORM=\"cm4io_nfs\"" /root/build.info ; then
+                    mesh_service
+            fi
+    else
     /bin/bash "$BRIDGE_SETTINGS"
+    fi
```
joenpera, line 436: Why we are starting mesh_service two times in one specific case?

joenpera (changes requested):

modules/utils/docker/entrypoint.sh
govindsi: ahh it was typo, fixed. But probably will investigate why ethX is not transparent in NFS boot. Then we can finalize on this change.

PR #310, opened 2023-09-14
Enable QoS-OLSR to run on the eth bridge

francislin12:
We can summarize our pull request as : 
1) We can switch between Batman-adv and QoS-OLSR
2)  QoS-OLSR can run on the bridge mode 
3) we assign an IP for QoS-OLSR and dynamic IP for dhcpd


modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,111 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
+}
+
+
+
+# Bridge ethernet and Mesh for Batman-avd and routing for QoS-OLSR
+if [ "$algo" = "olsr" ]; then
+  brctl addif br-lan "$eth_port"
+  IPV6_PREFIX=$( echo fd`dd if=/dev/urandom bs=7 count=1 status=none | xxd -p` | sed 's/\(....\)/\1:/g' )
+  ip addr add $SUBNET.1/24 dev br-lan && ip addr add $IPV6_PREFIX:1/64 dev br-lan
+  create_olsrd_config "wlp1s0" "$SUBNET" "$IPV6_PREFIX"
```
mbaddeley, line 125: We should find the interface name (i.e., wlxxxx) rather than hardcode it. Like in mcc_settings.sh. Same does for all other hardcoded wlp1s0 references in these changes.

modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,111 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
+  wlp1s0_ip="192.168.11."$((16#$ip_random))
+  ifconfig br-lan up
+  ifconfig wlp1s0 "$wlp1s0_ip" 
+  iptables -A FORWARD --in-interface $mesh_if -j ACCEPT
+  iptables -A FORWARD -i br-lan -o $mesh_if -j ACCEPT 
+  iptables -A FORWARD -o br-lan -i $mesh_if -j ACCEPT
+  killall olsrd 2>/dev/null	
+  (qos-olsrd -i wlp1s0 -d 0)&
+  ifconfig "$mesh_if" mtu 1500
+else
```
mbaddeley, line 139: the "else" block here should try and just use the code that is already there so there are fewer changes. In this case the lines are actually the same (just in a different order), but in future try to change as little as possible to make things easier to review.

modules/utils/docker/bridge_settings.sh
```diff
@@ -32,7 +32,7 @@ br_lan_ip="192.168.1."$((16#$ip_random))
 # bridge_settings
 echo "Running bridge_settings"
 brctl addbr br-lan
-# Bridge ethernet and Mesh
+# Selecting the interfaces
```
mbaddeley, line 35: There is no reason to change this.

modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,111 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
-ifconfig bat0 mtu 1500
-ifconfig br-lan mtu 1500
-
-brctl addif br-lan bat0 $eth_port
-echo $br_lan_ip
-ifconfig br-lan $br_lan_ip netmask "255.255.255.0"
-ifconfig br-lan up
-echo
-ifconfig br-lan
+create_dhcpd_config()
```
mbaddeley, line 53: Generally we now have duplicates for these functions (create_dhcpd_config(), create_olsrd_config(), and create_radvd_config()) with mcc_settings.sh. I would like to see an "qos-olsr_config.sh" where these now go and then we have a "main()" function which calls them. You can then just call that bash script from BOTH mcc_settings.sh and bridge_settings.sh

modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,111 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
+  ifconfig bat0 mtu 1500
+  ifconfig br-lan mtu 1500
+  echo "$br_lan_ip"
+  ifconfig br-lan "$br_lan_ip" netmask "255.255.255.0"
+  ifconfig br-lan up
+  echo
+  ifconfig br-lan
+fi
+#Automatically assign the IP address on the brlan for the other interface 
+dhcpd -cf  /etc/dhcp/dhcpd.conf br-lan
```
mbaddeley, line 153: Should this dhcpd command go here or in the qos-olsr "if" block?

mbaddeley (changes requested):

modules/utils/docker/bridge_settings.sh
TIISR: Hi! Names like wlp1s0 shouldn't be hardcoded indeed as they are likely to change when we swap the RPiCM4 for a SalukiPi for example since wlp1s0 means something like "WiFi card on PCI(e) port 1.0", but programmatically _finding_ the interface name might also be risky: maybe safer to have it as a configured variable name if that is not the case already?

Also I might have written that myself but upon careful consideration there should not be a `&&` preventing IPv6 configuration if IPv4 failed, especially since IPv4 is the more problematic one: I would recommend writing the two `ip` commands on separate lines so that if either fails then hopefully we at least still have the other :)

modules/utils/docker/bridge_settings.sh
TIISR: Fully agree :) but there are many other issues with HOS1.0 shell scripts anyway, would it make more sense to go straight to Python with unit testing etc? We'll likely need to share these create_dhcpd_config() and created_radvd_config() across even more modes so it might make sense to have them as dedicated classes in Python for example, or at least separate dhcpd.sh and radvd.sh files... These two being only two examples, there are probably even more topics (hostapd? wpa_supplicant?) that could benefit from being refactored into separate files/classes :)

modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,111 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
+  SUBNET="$1"
+
+  cat > /etc/dhcp/dhcpd.conf <<- EOF
+  default-lease-time 600;
+  max-lease-time 7200;
+  ddns-update-style none;
+  authoritative;
+
+  subnet $SUBNET.0 netmask 255.255.255.0 {
+    range $SUBNET.100 $SUBNET.199;
```
TIISR, line 64: Should be `range $SUBNET.2 $SUBNET.254;` 

The .100 to .199 range sounds like a leftover from testing when other devices were already statically configured as .2, .3, etc :)

modules/utils/docker/bridge_settings.sh
francislin12: it is for batman and qos-olsr

joenpera: Previously we didn't dhcp in batman mesh.  Is this a desired change?

TIISR (commented):
Sorry many of my comments turned out to be "Pending"

joenpera (changes requested):
PR is pointing to main

modules/utils/docker/bridge_settings.sh
mbaddeley: Eventually yes. For MS v2.0 I think this is a change we should avoid. We have a meeting about it this afternoon.

modules/utils/docker/mcc_settings.sh
```diff
@@ -49,6 +48,16 @@ calculate_wifi_channel()
     fi
 }
 
+create_ap_config()
```
mbaddeley, line 51: create_ap_config() is never called. We should remove in this case.

mbaddeley (changes requested):

mbaddeley:
@francislin12 can you please address both my changes and the requested change from @joenpera?

modules/utils/docker/bridge_settings.sh
mbaddeley: > 1. Is line 136 meant to kill the original OLSR daemon? If so, why do we have both installed?
> 2. Is line 136 (also) meant to kill the QoS-OLSR daemon? If so, it needs to be killall qos-olsrd or pkill olsrd :)
> 3. Is it ok that we are killing absolutely all instances instead of a specific PID?
> 4. According to olsrd manpage -d 0 makes it run in the background so line 137 should need neither () nor &
I believe this is just a hangover from before the daemon name change. @francislin12 please fix.

modules/utils/docker/bridge_settings.sh
mbaddeley: Sure, can you create an issue in github? We can then track it and pull it in later.

modules/utils/docker/bridge_settings.sh
mbaddeley: I see on line 103 we are still calling radvd

> radvd -C /etc/radvd.conf -d 0

This will be called in batman as well as olsr. Based on what you said @TIISR, should this be moved to OLSR only?

modules/utils/docker/bridge_settings.sh
mbaddeley: We now seem to have $WIFI_INTERFACE but I have no idea where this var is coming from as it's not in the default mesh.conf. @francislin12 you will need to create another PR for tc_distro which creates the default mesh.conf

modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,111 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
+
+# Bridge ethernet and Mesh for Batman-avd and routing for QoS-OLSR
+if [ "$algo" = "olsr" ]; then
+  brctl addif br-lan "$eth_port"
+  IPV6_PREFIX=$( echo fd`dd if=/dev/urandom bs=7 count=1 status=none | xxd -p` | sed 's/\(....\)/\1:/g' )
+  ip addr add $SUBNET.1/24 dev br-lan && ip addr add $IPV6_PREFIX:1/64 dev br-lan
+  create_olsrd_config "wlp1s0" "$SUBNET" "$IPV6_PREFIX"
+  create_dhcpd_config "$SUBNET"
+  dhcpd -f br-lan
+  create_radvd_config "$IPV6_PREFIX"
```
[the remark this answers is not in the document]
mbaddeley: Given this, should we not remove all IPv6 references for now? Seeing as how it is not being (and cannot be) used at the moment.

modules/utils/docker/qos_olsrd_conf.sh
```diff
@@ -0,0 +1,89 @@
...
+}
+
+create_olsrd_config()
+{
+  wifidev="$1"
+  SUBNET="$2"
+  cat > /etc/olsrd/olsrd.conf <<- EOF
+  LinkQualityFishEye   0
+
+  Interface "$WIFI_INTERFACE"
```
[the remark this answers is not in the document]
joenpera: Sounds that we don't have tested content in the review.

joenpera (changes requested):
PR still pointing to main.

modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,111 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
+  br_lan_ip="$SUBNET."$((16#$ip_random))	
+  wlp1s0_ip="192.168.11."$((16#$ip_random))
```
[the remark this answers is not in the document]
mbaddeley: Agreed, but this is not the scope of this PR. Can you create a GitHub issue for this?

modules/utils/docker/qos_olsrd_conf.sh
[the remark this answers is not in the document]
mbaddeley: Agreed. @francislin12 when you test you need to make sure you don't manually add anything. Otherwise this will not work out-of-the-box for others.

modules/utils/docker/bridge_settings.sh
```diff
@@ -50,20 +50,55 @@ if [ -f "$COMMS_PCB_VERSION_FILE" ]; then
...
-route add -net 192.168.1.0 gw $br_lan_ip netmask 255.255.255.0 dev br-lan
-iptables -A FORWARD --in-interface bat0 -j ACCEPT
-iptables --table nat -A POSTROUTING --out-interface $br_lan_ip -j MASQUERADE
\ No newline at end of file
+
+
+# Bridge ethernet and Mesh for Batman-avd and routing for QoS-OLSR
+if [ "$algo" = "olsr" ]; then
+  brctl addif br-lan "$eth_port"
+  IPV6_PREFIX=$( echo fd`dd if=/dev/urandom bs=7 count=1 status=none | xxd -p` | sed 's/\(....\)/\1:/g' )
```
joenpera, line 59: Not using SLAAC?

modules/utils/docker/qos_olsrd_conf.sh
```diff
@@ -0,0 +1,89 @@
+
+#! /bin/bash
+
+
+create_dhcpd_config()
```
joenpera, line 5: How DHCP spoofing is prevented?

modules/utils/docker/bridge_settings.sh
mbaddeley: I think @TIISR also had questions about this - for now let's leave it and we are working on another PR to change everything to SLAAC

modules/utils/docker/qos_olsrd_conf.sh
mbaddeley: It's not, I guess? However I believe with the move to SLAAC (will be another PR) then we will no longer be using DHCP. @TIISR am I correct?

mbaddeley:
UPDATE: @francislin12 is building a new image with the default mesh.conf updated to include WIFI_INTERFACE. He will then run tests on QoS-OLSR and BATMAN to ensure this works and, if so, I suggest we accept this PR and move the comments on IPv6 to a new PR we will be creating to fully implement SLAAC. Thoughts?

mbaddeley:
Superseded by #323

PR #317, opened 2023-10-04
Mesh com 2.0

SelinaTII:
PHYCRA changes


dania-tii (approved):

joenpera:
Just a comment using python and not about the actual functionality:

- when using python classes, define which functions should be public and which private functions. 
- static type definitions --> easier to see what function returns and type for the parameters --> less bugs 
- static type definitions --> later we can try to use compilers to boost up python code execution.

martin-tii (approved):

PR #319, opened 2023-10-09
Multiradio support

joenpera:
Implements messaging interface for Multiradio control.

Jira-Id: SPSI-43


common/scripts/mesh-helper.sh
```diff
@@ -43,21 +43,34 @@ EOF
   fi
 }
 
-generate_br_lan_ip() {
+generate_bridge_ip() {
```
TIISR, line 46: Please do not remove the understanding that this function is only for br-lan as there will be quite a few other bridges :)

common/scripts/mesh-helper.sh
```diff
@@ -43,21 +43,34 @@ EOF
...
   local mesh_if_mac
-  mesh_if_mac="$(ip -brief link | grep "$MESH_VIF" | awk '{print $3; exit}')"
+
+  bridge_name=$(echo "$BRIDGE" | cut -d' ' -f1)
```
TIISR, line 49: If this function is not for br-lan only but for any "bridge" then please pass the interface name as argument to the function, like so for example: `BRIDGE_NAME="$1"` :)

common/scripts/mesh-helper.sh
```diff
@@ -43,21 +43,34 @@ EOF
...
+
+  mesh_if_mac="$(ip -brief link | grep "${id0_MESH_VIF}" | awk '{print $3; exit}')"
```
TIISR, line 51: What is $id0_MESH_VIF? If it is an interface name (like "wlan0" as opposed to an index like "1") then it is much faster and more reliable to `MESH_IF_MAC=$(cat /sys/class/net/$id0_MESH_VIF/address)` :)

mbaddeley (approved):

common/scripts/mesh-helper.sh
```diff
@@ -43,21 +43,34 @@ EOF
...
+  if [ -z "$mesh_if_mac" ]; then
+      echo "generate_bridge_ip: MAC not found for id0_MESH_VIF! Configuration error?" > /dev/kmsg
+      mesh_if_mac="$(ip -brief link | grep "eth0" | awk '{print $3; exit}')"
+  fi
   local ip_random
   ip_random="$(echo "$mesh_if_mac" | cut -b 16-17)"
```
TIISR, line 57: As reported before but maybe not noticed yet: **IPv4 CANNOT BE SELF-CONFIGURED** :)

For example:

- If the MAC address ends in 00 then the bridge will try to use 192.168.1.0 i.e. the reserved network address in this /24
- If the MAC address ends in ff then the bridge will try to use 192.168.1.255 i.e. the reserved broadcast address in this /24
- It takes only 19 CommsDevices for "two of them picked the same IP" to happen more often than not.

**Please stop using IPv4** wherever possible :) For example here
- the interface most likely already has an IPv6 LLA (Link Local Address, the kind that starts with fe80)
- which you can change to one derived from one of the CommsDevice's physical MAC addresses if that is not what it did already
- or enable Duplicate Address Detection (not the best)
- or give it an IPv6 Unique Local Address with the first half `fd[something chosen randomly but once forever]` and the second half an EUI-64 from one of the CommsDevice's physical MAC addresses for example 00:20:91:ab:cd:ef would become 0220:91ff:fe:cdef (simple algorithm that inserts ff:fe in the middle and flips the Locally Administered bit).

common/scripts/mesh-helper.sh
```diff
@@ -43,21 +43,34 @@ EOF
...
-  br_lan_ip="192.168.1."$((16#$ip_random))
+  bridge_ip="192.168.1."$((16#$ip_random))
+
+  # legacy support
+  br_lan_ip=$bridge_ip
 }
 
 source_configuration() {
-  if [ -f "/opt/mesh.conf" ]; then
+  # $1: index of the mesh.conf file to be used
```
TIISR, line 65: If I may recommend as a coding style requirement or something: since shell script functions are not clear at all about which arguments they take if any, please ensure that at least the first few lines of the function body show them, for example:
`MESH_CONF_FILE_INDEX="$1"`
`WHAT_THE_2ND_ARG_IS="$2"`
`WHAT_THE_3RD_ARG_IS="$3"`
and so on :)

common/scripts/mesh-helper.sh
```diff
@@ -43,21 +43,34 @@ EOF
...
+
+  if [ -f "/opt/${1}_mesh.conf" ]; then
     # Calculate hash for mesh.conf file
```
TIISR, line 68: Please explain why, I for one struggle to understand what this is about :)

If this is a way to help parse the configuration then maybe we need to brainstorm a better way such as breaking it into a conf.d/ directory for example as it more and more common practice.

Otherwise checking if a file changed is already doable better with `diff` already (e.g. a hash will be completely different if only a newline was added for example, while with `diff` such insignificant changes can be ignored) or if we really want to detect any sort of byte difference then these steps can be replaced simply with the `cmp` command :)

TIISR (commented):
First batch, will try to continue later

modules/sc-mesh-secure-deployment/src/nats/initd/S9011sNatsMesh
```diff
@@ -10,35 +10,62 @@
...
     return "$status"
 }
 stop() {
     printf 'Stopping %s: ' "$DAEMON"
-    NEW_PID=$(ps ax | grep -E 'wpa_supplicant\-11s' | awk {'print $1'})
-    kill -9 $NEW_PID
-    rm -fr /var/run/wpa_supplicant/* 2>/dev/null
-    ifconfig "$MESH_VIF" down 2>/dev/null
+    kill -9 "$(cat "$PIDFILE")"
+    kill -9 "$(ps ax | grep -E "[w]pa_supplicant\-11s_$RADIO_INDEX" | awk {'print $1'})"
```
TIISR, line 66: Please do not kill -9 (SIGKILL) right away, the correct way is to send SIGTERM first to give programs a chance to finish up and/or exit cleanly while we wait for a bit, then SIGKILL them only if they haven't.

This is another reason to use a proper daemon manager like SystemD to not have to write lots of little paragraphs here and there reinventing the wheel and probably not doing it as well as a modern init system :)

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_radio_on.py
```diff
@@ -8,7 +8,7 @@ async def main():
     # Connect to NATS!
     nc = await client.connect_nats()
 
-    cmd_dict = {"api_version": 1, "cmd": "UP"}
+    cmd_dict = {"api_version": 1, "cmd": "UP", "radio_index": "0"}
```
TIISR, line 11: Above I can guess from the script names (...hostapd.py, ...wpa.py, etc) that "radio_index" only applies to WiFi interfaces but what about here: would "cli_radio_on.py" be called to turn OSF on for example? :)

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
```diff
@@ -3,27 +3,80 @@
...
-                "role": f"{config.MODULE_ROLE}"}
+    cmd_dict = {
+        "api_version": 1,
+        "role": f"{config.MODULE_ROLE}",  # sleeve, drone, gcs
+        "radios": [
+            {
+                "radio_index": "0",
+                "ssid": "test_mesh2",
+                "key": "1234567890",
+                "ap_mac": "00:11:22:33:44:55",
```
TIISR, line 17: I see it was in the previous version of the file already but if we are writing this again then maybe it is the time to ask: what is ap_mac for? Can it be removed if not used?

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
```diff
@@ -3,27 +3,80 @@
...
+                "country": "US",  # all radios must have the same country
+                "frequency": "2412",
+                "frequency_mcc": "2412",  # multiradio not supporting
+                "routing": "batman-adv",
+                "priority": "long_range",
+                "ip": "10.20.15.3",
```
TIISR, line 23: Please remove all IPv4 or at least leave it empty to avoid the happened-several-times-already situations where we debug and debug and debug only to find out that several devices are using 10.20.15.3 etc :)

martin-tii (approved):

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
```diff
@@ -3,27 +3,80 @@
...
+                "radio_index": "1",
+                "ssid": "test_mesh",
+                "key": "1234567890",
+                "ap_mac": "00:11:22:33:44:55",
+                "country": "US",  # all radios must have the same country
+                "frequency": "5220",
+                "frequency_mcc": "2412",  # multiradio not supporting
+                "routing": "batman-adv",
+                "priority": "long_range",
+                "ip": "10.20.15.3",
```
TIISR, line 40: Conflicts with radio_index 0, please remove all IPv4.

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
```diff
@@ -3,27 +3,80 @@
...
+                "radio_index": "2",
+                "ssid": "test_mesh3",
+                "key": "1234567890",
+                "ap_mac": "00:11:22:33:44:55",
+                "country": "US",  # all radios must have the same country
+                "frequency": "5190",
+                "frequency_mcc": "2412",  # multiradio not supporting
+                "routing": "batman-adv",
+                "priority": "long_range",
+                "ip": "10.20.15.3",
```
TIISR, line 57: Conflicts with radio_index 0 and 1, please remove all IPv4.

common/scripts/mesh-helper.sh
joenpera: I generalized the name, because the bridge name is configurable.

common/scripts/mesh-helper.sh
joenpera: It is not a command line argument.  Configuration sourced to file.

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
```diff
@@ -3,27 +3,80 @@
...
+                "subnet": "255.255.255.0",
+                "tx_power": "30",
+                "mode": "halow",  # ap+mesh_scc, mesh, halow
+                "mesh_vif": "halow1",
+                "batman_iface": "bat0",
```
TIISR, line 62: Selina's work still needs to be merged into this CommsController but just in case it helps to mention now, in short:

1. Each interface (11s WiFi, STA WiFi, RJ45, USB, etc) such as halow1 here will be given to the Network Access Control logic
2. NAC will detect peers (with help from MESH-PEER-CONNECTED notifications on the wpa_supplicant control interface wherever possible) and setup a "lower MACsec" interface for each one of them, for example lms002091abcdef to for the secure channel between here and the 00:20:91:ab:cd:ef interface on the other side.
3. Each new lms* interface gets added to the lower MACsec bridge called br-lms for example, which filters (a lot like "client isolation" in a WiFi AP) to allow traffic only between each lms* and the veth uplink explained next, as opposed to allowing traffic between two lms* interfaces which would create a network loop otherwise.
4. Adding batman-adv directly on top on br-lms (not to be confused with how so far we have been adding a bridge on top of a batman-adv interface like bat0 as one of the downlinks of br-lan) would make it a bit difficult to write the filtering rules from the previous point, so we add a "veth" imaginary cable: one side is given to the switch like `brctrl addif br-lms veth-foo` and the other side is given to batman-adv like `batctl [meshif bat0] add if veth-bar`.
5. Steps 2, 3 & 4 are repeated for what we call the "upper" part which results in things like ums002091abcdef, br-ums, bat1, etc.
6. NATS Provisioning etc (e.g. OS updates in the future) can use bat1.

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_visual_subscribe.py
```diff
@@ -24,13 +24,12 @@ async def subscribe_handler(msg):
...
-        print("Received a message on '{subject} {reply}': {data}".format(
-          subject=subject, reply=reply, data=data))
+        print(f"Received a message on '{subject} {reply}': {data}")
 
     try:
-        await client.connect(nc, reconnected_cb=reconnected_cb,
+        await client.connect(nc, recon_cb=reconnected_cb,
                              closed_cb=closed_cb,
-                             max_reconnect_attempts=-1)
+                             max_recon_attempts=-1)
```
TIISR, line 32: While there was room for improvement in the "reconnect" vs "reconnected" naming, it might be worth to not shorten so much down to "recon" which is likely to get confused with "reconnaissance" :)

modules/sc-mesh-secure-deployment/src/nats/scripts/config.py
```diff
@@ -8,12 +8,12 @@
     print("No identity found!!!!!!!!!!!!!!!!!!!!!!!!")
     print("Please run _cli_command_get_identity.py to get the identity (creates identity.py)")
 
-MODULE_IP = "10.10.10.2"  # or 192.168.1.x - brlan ip address
+MODULE_IP = "10.10.20.2"  # or 192.168.1.x - brlan ip address
```
TIISR, line 11: Please remove all IPv4

modules/sc-mesh-secure-deployment/src/nats/scripts/identity.py
```diff
@@ -0,0 +1 @@
+MODULE_IDENTITY="00301a4fc7e4100000001e8cac33"
```
TIISR, line 1: I guess that's what it is for now but just in case it helps to mention: needs to be replaced by birth certs as soon as possible :)

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
```diff
@@ -18,21 +19,143 @@ def test_handle_mesh_settings(self):
...
+                    "radio_index": "0",
+                    "ssid": "test_mesh2",
+                    "key": "1234567890",
+                    "ap_mac": "00:11:22:33:44:55",
+                    "country": "US",  # all radios must have the same country
+                    "frequency": "2412",
+                    "frequency_mcc": "2412",  # multiradio not supporting
+                    "routing": "batman-adv",
+                    "priority": "long_range",
+                    "ip": "10.20.15.3",
```
TIISR, line 43: Please remove all IPv4

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
```diff
@@ -18,21 +19,143 @@ def test_handle_mesh_settings(self):
...
+                    "radio_index": "1",
+                    "ssid": "test_mesh",
+                    "key": "1234567890",
+                    "ap_mac": "00:11:22:33:44:55",
+                    "country": "US",  # all radios must have the same country
+                    "frequency": "5220",
+                    "frequency_mcc": "2412",  # multiradio not supporting
+                    "routing": "batman-adv",
+                    "priority": "long_range",
+                    "ip": "10.20.15.3",
```
TIISR, line 60: Please remove all IPv4

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
```diff
@@ -18,21 +19,143 @@ def test_handle_mesh_settings(self):
...
+                    "radio_index": "2",
+                    "ssid": "test_mesh3",
+                    "key": "1234567890",
+                    "ap_mac": "00:11:22:33:44:55",
+                    "country": "US",  # all radios must have the same country
+                    "frequency": "5190",
+                    "frequency_mcc": "2412",  # multiradio not supporting
+                    "routing": "batman-adv",
+                    "priority": "long_range",
+                    "ip": "10.20.15.3",
```
TIISR, line 77: Please remove all IPv4

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
```diff
@@ -18,21 +19,143 @@ def test_handle_mesh_settings(self):
...
+                    "radio_index": "0",
+                    "ssid": "test_mesh2",
+                    "key": "1234567890",
+                    "ap_mac": "00:11:22:33:44:55",
+                    "country": "US",  # all radios must have the same country
+                    "frequency": "2400",
+                    "frequency_mcc": "2412",  # multiradio not supporting
+                    "routing": "batman-adv",
+                    "priority": "long_range",
+                    "ip": "10.20.15.3",
```
TIISR, line 113: Please remove all IPv4

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
```diff
@@ -18,21 +19,143 @@ def test_handle_mesh_settings(self):
...
+                    "radio_index": "1",
+                    "ssid": "test_mesh",
+                    "key": "1234567890",
+                    "ap_mac": "00:11:22:33:44:55",
+                    "country": "US",  # all radios must have the same country
+                    "frequency": "5220",
+                    "frequency_mcc": "2412",  # multiradio not supporting
+                    "routing": "batman",
+                    "priority": "long_range",
+                    "ip": "10.20.15.3",
```
TIISR, line 130: Please remove all IPv4

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
```diff
@@ -18,21 +19,143 @@ def test_handle_mesh_settings(self):
...
+                    "radio_index": "2",
+                    "ssid": "test_mesh3",
+                    "key": "123456 7890",
+                    "ap_mac": "00:11:22:33:44:55",
+                    "country": "US",  # all radios must have the same country
+                    "frequency": "5190",
+                    "frequency_mcc": "2412",  # multiradio not supporting
+                    "routing": "None",
+                    "priority": "looong_range",
+                    "ip": "10.20.15.3",
```
TIISR, line 147: Please remove all IPv4

common/scripts/mesh-helper.sh
joenpera: This is not a valid comment for multiradio implementation.  We are re-using a code which was done much earlier.   IPv4 to IPv6 change was not in a scope of multiradio work.

modules/utils/docker/entrypoint_nats.sh
```diff
@@ -18,20 +18,42 @@ else
...
   fi
 
+  # Do not continue in case halow init has not finished
+  while ps aux | grep [i]nit_halow > /dev/null; do
+      sleep 1
+  done
+
   echo "set bridge ip"
-  generate_br_lan_ip
+  generate_bridge_ip
```
TIISR, line 27: As mentioned above: there will be quite a few bridges, if this is only for br-lan then let's keep it in the name :)

TIISR (commented):
Written in a hurry, apologies if anything sounds harsh as it is not at all how I mean all this :)

modules/sc-mesh-secure-deployment/src/nats/initd/S9011sNatsMesh
```diff
@@ -10,35 +10,62 @@
...
 [ -r "/etc/default/$DAEMON" ] && . "/etc/default/$DAEMON"
 
 start() {
+    if mkdir $SYNC_LOCK 2>/dev/null; then
+       echo "I have a lock $RADIO_INDEX PID $$"
+    else
+       echo "Instance is already running. Waiting..."
+       # Wait for the lock directory to be removed, indicating the previous instance finished
+       while [ -d "$SYNC_LOCK" ]; do
+         sleep $(( (RANDOM % 5) + 1 ))
```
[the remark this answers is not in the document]
joenpera: Thanks.  Will be fixed.

modules/sc-mesh-secure-deployment/src/nats/initd/S9011sNatsMesh
joenpera: I know that we have limitations with init.d and currently we need to live with that.

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_radio_on.py
joenpera: OSF is not yet part of multiradio.  It was agreed when this work was ongoing.

common/scripts/mesh-helper.sh
TIISR: Sorry I am new to PR comments, I was hoping this is an opportunity to clean up the "going to waste hours in debug" broken IPv4 code.

modules/utils/docker/entrypoint_nats.sh
joenpera: Bridge name is in config parameters.

common/scripts/mesh-helper.sh
TIISR: Then please call it multiradiobridge for example, not just "bridge" as that introduces confusion with other bridges :)

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
joenpera: ap_mac is used for wpa_supplicant bssid.

modules/sc-mesh-secure-deployment/src/nats/initd/S9011sNatsMesh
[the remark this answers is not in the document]
joenpera: It seem's as it is arithmetic function $ is not needed.

common/scripts/mesh-helper.sh
joenpera: It can be used also with one radio, that is the reason to generalize it to bridge.

common/scripts/mesh-helper.sh
joenpera: Some pseudo lines added.  Maybe that helps.   Hash is for to detect modified file without using correct interfaces.
```
source_configuration() {
  # $1: radio index of the mesh.conf file to be used

  # if we have radio specific config written into device over NATS
  if [ -f "/opt/${1}_mesh.conf" ]; then
    if hash ok; then
      source /opt/"${1}"_mesh.conf
    else
      # use default mesh, if hash is not ok
      if default available (generated in first boot); then
        source /opt/mesh_default.conf
      else
        echo "no valid mesh configuration found!"
		exit 1
      fi
    fi
  elif if we don't have radio specific file written --> use default; then
      source /opt/mesh_default.conf
  else
      echo "no valid mesh configuration found!"
      exit 1
  fi
}
```

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 remova

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 removal

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 removal

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 removal

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 removal

common/scripts/mesh-helper.sh
joenpera: This is not enough?
`# $1: index of the mesh.conf file to be used`

modules/sc-mesh-secure-deployment/src/nats/scripts/config.py
joenpera: config.py is the configuration file for testing scripts.  As we have IPv4 currently in multiradio --> it is needed.   Cleanup will be done when IPv6 will be introduced.

modules/sc-mesh-secure-deployment/src/nats/scripts/identity.py
joenpera: this file needs to be removed.  It is auto-generated when asking identity.   Identity is needed for communication with NATS servers.  Something similar needs to be on place also after birth certs.

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 removal

modules/sc-mesh-secure-deployment/src/nats/tests/test_settings.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 removal

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
joenpera: Yes, but with separate PR.  This is for multiradio implementation.

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
joenpera: I think coming clean-up with ipv6 can include all ipv4 removal

common/scripts/mesh-helper.sh
joenpera: fixed in b67d5f9e3c84fcec3732cb04a35297e5acbe08f4

common/scripts/mesh-helper.sh
joenpera: fixed in https://github.com/tiiuae/mesh_com/commit/b67d5f9e3c84fcec3732cb04a35297e5acbe08f4

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_visual_subscribe.py
joenpera: fixed in https://github.com/tiiuae/mesh_com/commit/b67d5f9e3c84fcec3732cb04a35297e5acbe08f4

modules/sc-mesh-secure-deployment/src/nats/scripts/identity.py
joenpera: fixed in https://github.com/tiiuae/mesh_com/commit/b67d5f9e3c84fcec3732cb04a35297e5acbe08f4

File removed as automatically generated.   Birth cert things were not in the scope.

pemanty (approved):

common/scripts/mesh-helper.sh
TIISR: Fair enough but I would still call it something clear like LAN_BRIDGE_NAME then for example, especially if it is something coming all the way from mesh.conf and/or NATS server, where I for one would be definitely confused by a mysterious variable only called BRIDGE 🙂

common/scripts/mesh-helper.sh
TIISR: Was maybe another block supposed to be marked as resolved? I do see several changes (thank you) in the mentioned commit but I can't seem to find a change related to this? 🤔

common/scripts/mesh-helper.sh
TIISR: ok but what do we do for now about this not being able to work if someone ends up having a MAC address ending in 00 or FF for example? We blame hardware and send another device? 🙂

common/scripts/mesh-helper.sh
TIISR: Your comment line is not bad in itself, I saw it as the very first line of the function indeed, but later in the function it would still be better to have clear variable names than having to remember "What was $1 again?" 🙂

It would be also safer as it's much harder to typo $WHAT_I_MEANT into $SOMETHING_COMPLETELY_DIFFERENT than it is to typo $1 into $2

modules/sc-mesh-secure-deployment/src/nats/initd/S9011sNatsMesh
TIISR: Still need to issue SIGTERM then wait then SIGKILL if necessary 🙂

Otherwise not giving a chance for programs to exit gracefully is likely to create all sorts of unreliability like maybe you won't be able to start another wpa_supplicant or hostapd because the previous one left the interface in a busy state, or at least not without unloading+reloading the driver, maybe rebooting the whole OS or who knows what, much more complicated than just not murdering programs all the time 🙂

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_radio_on.py
TIISR: Maybe I did not express myself very well but ok if I hopefully understand correctly now that OSF will be added to this in the future then the name does not need to be WiFi related 🙂

common/scripts/mesh-helper.sh
joenpera: I suppose IPv6 should fix this or is there bottleneck in slaac design?

modules/utils/docker/entrypoint_nats.sh
TIISR: How about generate_lan_bridge_ip or something clearer like that? 🙂

TIISR (commented):
Not sure if my comments get stuck as Pending again unless I do this

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
joenpera: It's named as bssid in wpa supplicant.  Again this is not related to multiradio implementation.  It has been almost 3 years like and tested to be working with MP.

govindsi (approved):

common/scripts/mesh-helper.sh
joenpera: Minor thing, but I will fix this.

joenpera: Fixed in https://github.com/tiiuae/mesh_com/pull/319/commits/d730ee831cee33375488ea0117601abd4c7f81da

modules/utils/docker/entrypoint_nats.sh
joenpera: fixed https://github.com/tiiuae/mesh_com/pull/319/commits/5353cf47e111210414924b59e97faec42b8f29cc

modules/sc-mesh-secure-deployment/src/nats/initd/S9011sNatsMesh
joenpera: Not going to fix this now.

common/scripts/mesh-helper.sh
joenpera: Resolving this as IPv6 was not in scope.

modules/sc-mesh-secure-deployment/src/nats/scripts/config.py
mbaddeley: Sure, this is an issue but as with a lot of stuff at the moment, I think this is a "known issue" that needs to be logged as a separate ticket in the MS Dev Jira and picked up as a separate PR.

joenpera: That connection is for identity request.  You are just connecting to device which is connected with Ethernet cable.  Later, if you need to send messages to other device there should be used bridge IP and 10.10.20.2 as a gateway.

PR #334, opened 2023-11-15
Pytorch package installation

dania-tii:
Installing torch-1.13.1
- This package is required for jamming avoidance and LDPI (IDS)
- It cannot be installed via buildroot as the source code is not available

### DIFF
--- a/modules/utils/docker/entrypoint_nats.sh
+++ b/modules/utils/docker/entrypoint_nats.sh
@@ -74,9 +74,12 @@ else
   /opt/S90comms_controller start
 
   # Install python package "torch"
-  echo "installing packages"
-  chmod +x /opt/mesh_com/modules/utils/package/install_dependencies.sh
-  /opt/mesh_com/modules/utils/package/install_dependencies.sh
+  os_type=$(uname -s)
+  if [ "$os_type" = "Linux" ]; then
+      echo "installing packages"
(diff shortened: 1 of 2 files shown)


martin-tii (approved):

joenpera (changes requested):
includes only aarch64 installation packages.

dania-tii:
> includes only aarch64 installation packages.

I updated the code such that the installation of pytorch occurs for linux os. The IDS and jamming avoidance features, which require the pytorch package, for now are intended to be enabled only for CM4 platform.

joenpera:
I was thinking that should we have this installation process before entrypoint_nats.sh is called and actually in mesh_start.sh.   As now we are starting all processes and after that installing dependencies.

joenpera (changes requested):
Same was added already as comment:
I was thinking that should we have this installation process before entrypoint_nats.sh is called and actually in mesh_start.sh? As now we are starting all processes and after that installing dependencies.

dania-tii:
> Same was added already as comment: I was thinking that should we have this installation process before entrypoint_nats.sh is called and actually in mesh_start.sh? As now we are starting all processes and after that installing dependencies.

ok, I'll close this pr then and install the dependencies via mesh_start.sh

joenpera (approved):

PR #336, opened 2023-11-16
mdm_client support

joenpera:
Current:

- mDNS
- TLS support
- S90mdm_agent started in device boot

Jira-Id: MSS20-226, MSS20-227, MSS20-230


johan-juntunen (approved):
Looks good to me

andtokm (approved):
Thx you so much

modules/sc-mesh-secure-deployment/src/nats/src/validation.py
```diff
@@ -297,3 +298,18 @@ def validate_mptcp(mptcp: str) -> bool:
...
+    """
+    Validates a given slaac ifaces.
+    Returns True if the slaac ifaces are valid, False otherwise.
+    """
+    try:
+        for slaac_if in slaac.split():
+            if (not slaac_if.startswith("wl") and
+                    not slaac_if.startswith("halow") and
+                    not slaac_if.startswith("usb")):
+                return False
```
TIISR, line 312: Thank you for checking but I think this might create problems: for example we might want SLAAC on an RJ45 interface so a name like eth0, en[psx]* would get a False 🙂 And although this happens much more often in BSD than in Linux, I wonder if we might get yet another interface prefix just because some driver decided to call the interface blah0 or something 🤔

modules/utils/docker/entrypoint_nats.sh
```diff
@@ -13,16 +13,26 @@ if [ "$MSVERSION" != "nats" ]; then
...
+  #######################################
+  # Initialise default radios and IF    #
+  #######################################
+
   # Do not continue in case halow init has not finished
   while ps aux | grep [i]nit_halow > /dev/null; do
       sleep 1
   done
 
+  # TODO: remove as SLAAC and CBMA?
```
TIISR, line 35: Remove yes please: in general bridges are not even supposed to have an IP address 😉 In our case they already have an IPv6 Link-Local Address (fe80::...) anyway and when enabled, SLAAC is responsible for more addresses indeed.

TIISR (approved):

modules/sc-mesh-secure-deployment/src/nats/initd/S90mdm_agent
```diff
@@ -0,0 +1,68 @@
...
+
+DAEMON="comms_nats_controller.py"
+DIR="/opt/nats/"
+PIDFILE="/var/run/$DAEMON.pid"
+# shellcheck source=/dev/null
+[ -r "/etc/default/$DAEMON" ] && . "/etc/default/$DAEMON"
+LOG_FILE=/opt/mdm_agent.log
+
+KEY_FILE="/etc/ssl/private/csl1.local.key"
+CERT_FILE="/etc/ssl/certs/csl1.local.crt"
```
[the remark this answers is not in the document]
joenpera: Yes, these are for testing purposes only.   Temporary commit added in host side and will be removed once everything is on place.

modules/sc-mesh-secure-deployment/src/nats/scripts/cli_settings_request.py
```diff
@@ -29,6 +29,7 @@ async def main():
                 "mesh_vif": "wlp2s0",
                 "batman_iface": "bat0",
                 "bridge": "br-lan bat0 eth1 lan1 eth0 usb0",
+                "slaac": "usb0 wlp2s0", # SLAAC support for usb0 + mesh_vif
```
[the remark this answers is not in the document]
joenpera: script folder is for testing purposes and examples how to use NATS.io interface if needed (will be renamed to examples to give clarity).  SLAAC parameter is needed when radio interfaces are really initialized together with SLAAC, SLAAC is enabled for the radio only if we get parameters to do that.

modules/sc-mesh-secure-deployment/src/nats/src/validation.py
joenpera: Your suggestion is to remove validation for SLAAC parameter?

modules/utils/docker/entrypoint_nats.sh
joenpera: We need to wait CBMA to come in and then make other changes to support batman routing only.

joenpera: removed

modules/sc-mesh-secure-deployment/src/nats/src/validation.py
TIISR: I mean "this validation" of checking if the interface exists before launching the daemon with a typoed config (which might then be restarted many times until e.g. systemd realizes this is restarting too many times), and I just realized that the startsWith('wl') wouldn't catch the "wlna0" typo 🙂

joenpera: This is more like typo-catcher and needs improvements.

PR #337, opened 2023-11-17
CBMA code and test scripts

SelinaTII:
This is an initial PR for CBMA code and test scripts

- CBMA for an inteface can be setup by calling the cbma() function in https://github.com/tiiuae/mesh_com/blob/cbma/modules/sc-mesh-secure-deployment/src/2_0/features/cbma/setup_cbma.py
- The cbma() function calls setup_macsec() and setup_batman()
- An example of setting up cbma for wlp1s0 and eth1 is given in the main function of setup_cbma.py (steps to generate test certificates are also mentioned there)
- The code requires pyOpenSSL package (@dania-tii is integrating the build root package in tc_distro)
- The code has been tested for wlp1s0 and eth1 interfaces on CSLs (will perform further tests with other radios on comms module 2.0 devices)
- There are some improvements and bug fixes that will come up in later PRs


martin-tii (approved):

TIISR:
image

Context: Selina's code creates the lower/upper MACsec stacks (one interface per peer plus a bridge to aggregate them and give them as one to the batman-adv specified as parameter) visible in the diagram hereby attached.

The idea is for CommsController to  call Selina's code with configurations provided by MDM or by the default principle of bat0 as white and all physical(ish) interfaces as black.

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/macsec/macsec.py
```diff
@@ -0,0 +1,69 @@
...
+        except Exception as e:
+            self.logger.error(f'Error setting up {self.level} macsec tx channel with {client_mac}: {e}')
+            sys.exit(1)
+
+    def set_macsec_rx(self, client_mac, client_macsec_key, client_port):
+        # Adds a rx channel with client_mac, with key id = client mac without ":"
+        macsec_interface = self.get_macsec_interface_name(client_mac)
+        try:
+            subprocess.run(["ip", "macsec", "add", macsec_interface, "rx", "port", str(client_port), "address", client_mac], check=True)
+            subprocess.run(["ip", "macsec", "add", macsec_interface, "rx", "port", str(client_port), "address", client_mac, "sa", "0", "pn", "1", "on", "key", client_mac.replace(":", ""), client_macsec_key], check=True)
```
TIISR, line 40: I thought we agreed to have the peer with the smaller MAC address be the one setting Security Association 0 (alternating with 2 later when key refreshment support is added) and the one with the higher MAC setting SA 1 (alternating with 3 later) but here lines 40 and 27 both have '"sa", "0"' hardcoded so may I ask if the race condition (due to the risk of two key agreements being performed concurrently, one by ClientA & ServerB and one by ClientB & ServerA) was addressed differently?

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/macsec/macsec.py
```diff
@@ -0,0 +1,69 @@
...
+        self.used_ports = {} # client_mac: port
+        self.available_ports_lock = threading.Lock()
+        self.macsec_encryption = macsec_encryption  # Flag to set macsec encrypt on or off
+        self.logger = logger_instance.get_logger()
+
+    def set_macsec_tx(self, client_mac, my_macsec_key, my_port):
+        # Sets up macsec link and adds tx channel
+        macsec_interface = self.get_macsec_interface_name(client_mac)
+        try:
+            subprocess.run(["ip", "link", "add", "link", self.interface, macsec_interface, "type", "macsec", "port", str(my_port), "encrypt", self.macsec_encryption, "cipher", "gcm-aes-256"], check=True)
```
TIISR, line 26: Please add `send_sci off end_station on` to reduce the overhead from 32 to 24 :) This is possible because we create MACsec secure channels only between two MAC addresses ("end stations") as opposed to the more common use of it whereby the MACsec source might not be the same as the Ethernet source so this needs to be clarified with the Secure Channel Identifier.

TIISR (changes requested):

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/tools/utils.py
```diff
@@ -0,0 +1,346 @@
...
+
+def add_interface_to_batman(interface_to_add, batman_interface):
+    # Add interface to batman
+    try:
+        subprocess.run(["batctl", "meshif", batman_interface, "if", "add", interface_to_add], check=True)
+        logger.info(f'Added interface {interface_to_add} to {batman_interface}')
+    except Exception as e:
+        logger.error(f'Error adding interface {interface_to_add} to {batman_interface}: {e}')
+
+def add_interface_to_bridge(interface_to_add, bridge_interface):
```
TIISR, line 332: @pentestiing is double-checking but in case it helps to mention in the meantime, please add the Python equivalent of:

`ebtables --tables nat --append OUTPUT --out-interface $MACSEC_INTERFACE --destination ff:ff:ff:ff:ff:ff --jump dnat --to-destination $MACSEC_PEER_MAC_ADDRESS`

This is not mandatory but is good to avoid having batman-adv (at least the upper one) unnecessarily carrying broadcast frames to everyone when it's only the MACsec peer who can decrypt/verify🙂

And of course in a corresponding remove_interface_from_bridge function there should be the cleanup:

`ebtables --tables nat --delete OUTPUT --out-interface $MACSEC_INTERFACE --destination ff:ff:ff:ff:ff:ff --jump dnat --to-destination $MACSEC_PEER_MAC_ADDRESS`

TIISR (changes requested):

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authClient.py
```diff
@@ -0,0 +1,127 @@
...
+        self.secure_client_socket = None
+        self.logger = logger_instance.get_logger()
+        self.ca = ca_path
+        self.mymac = get_mac_addr(self.interface)
+        self.server_mac = server_mac
+        self.mua = mua
+
+    def establish_connection(self):
+        # Create an SSL context
+        context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
```
TIISR, line 35: Since we control both client and server we can use TLS 1.3 which is more secure and even faster 🙂

Python does not seem to make it easy though, as I can't find a ssl.PROTOCOL_TLSv1_3 in https://docs.python.org/3/library/ssl.html and according to only a bit of googling the ssl.PROTOCOL_TLSv* keywords are deprecated.

The way to use TLS1.3+ in Python is apparently:

```
context = ssl.SSLContext(ssl.PROTOCOL_TLS)
context.options |= (ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1 | ssl.OP_NO_TLSv1_2)
```

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authClient.py
```diff
@@ -0,0 +1,127 @@
...
+        while retries < MAX_RETRIES:
+            try:
+                if self.sslServerIP.startswith("fe80"):
+                    secureClientSocket.connect(
+                        (self.sslServerIP, self.sslServerPort, 0, socket.if_nametoindex(self.interface)))
+                else:
+                    secureClientSocket.connect((self.sslServerIP, self.sslServerPort))
+                break  # break out of loop if connection is successful
+            except ConnectionRefusedError:
+                retries += 1
```
TIISR, line 94: Possible infinite loop: `retries` gets incremented only in the case of ConnectionRefusedError but the try block can fail for a variety of reasons like TimeoutError, ConnectionAbortedError, ConnectionResetError... I guess here you should use `except Exception as e` like you did above🙂

TIISR (changes requested):

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
```diff
@@ -0,0 +1,136 @@
...
+        threading.Thread.__init__(self)
+        self.running = True
+        self.ipAddress = ip_address
+        self.port = port
+        self.CERT_PATH = cert_path
+        self.ca = ca_path
+        self.interface = interface
+        self.mymac = get_mac_addr(self.interface)
+        # Create the SSL context here and set it as an instance variable
+        self.context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
```
TIISR, line 28: TLS1.3 here too please 🙂

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
```diff
@@ -0,0 +1,136 @@
...
+            raise ValueError("Invalid IP address")
+
+        self.serverSocket.listen()
+        self.serverSocket.settimeout(99999)  # maybe we can remove timeout since server needs to be listening throughout
+        logger.info("Server listening")
+
+        while self.running and not self.mua.shutdown_event.is_set():
+            try:
+                client_connection, client_address = self.serverSocket.accept()
+                threading.Thread(target=self.handle_client, args=(client_connection, client_address)).start()
```
TIISR, line 103: Launching a new thread for any new connection (authenticated or not) is dangerous: an attacker can easily complete TCP handshakes all day while this program (or maybe even whole system) will be brought to its knees from too many threads; please start the client thread only after a client has successfully authenticated 🙂

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
```diff
@@ -0,0 +1,136 @@
...
+            self.serverSocket.bind((self.ipAddress, self.port))
+        elif is_ipv6(self.ipAddress):
+            self.serverSocket = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+            scope_id = socket.if_nametoindex(self.interface)
+            self.serverSocket.bind((self.ipAddress, int(self.port), 0, scope_id))
+        else:
+            raise ValueError("Invalid IP address")
+
+        self.serverSocket.listen()
+        self.serverSocket.settimeout(99999)  # maybe we can remove timeout since server needs to be listening throughout
```
TIISR, line 97: [Python experts please confirm]

If I hastily understand the documentation correctly, settimeout() is for specifying how many seconds we want to wait on each blocking socket call so 99999 is way too long as it would enable a client to keep us stuck for more than a day which is a denial of service in the main thread but also quite inconvenient in client threads...

Please change to 30s which is still a bit overkill but for safety in case an Upper MACsec client is dozens of hops away, and having something fairly small also gives us an opportunity to cleanup regularly in a "garbage collection" fashion when there is nothing else to do 🙂

TIISR (changes requested):

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
```diff
@@ -0,0 +1,136 @@
...
+            with self.client_auth_results_lock:
+                self.client_auth_results[client_address[0]] = auth
+            if auth:
+                with self.active_sockets_lock:
+                    self.active_sockets[client_address[0]] = secure_client_socket
+                self.mua.auth_pass(secure_client_socket=secure_client_socket, client_mac=client_mac)
+            else:
+                # Handle the case when authentication fails, maybe send an error message
+                self.mua.auth_fail(client_mac=client_mac)
+                secure_client_socket.send(b"Authentication failed.")
```
joenpera, line 70: send() - don't guarantee complete packet send.

https://docs.python.org/3/howto/sockets.html#using-a-socket

You need to have known message sizes.  Use sendall() to send complete packet and on receiver side you need to know how long packet is coming.

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/cert_generation/client_side.py
```diff
@@ -0,0 +1,101 @@
...
+    subprocess.run([CSR_SCRIPT_PATH, interface], check=True)
+    mac_address = get_mac_addr(interface)
+    csr_filename = glob.glob(f"macsec_{mac_address.replace(':', '')}.csr")[0]
+    crt = csr_filename.split(".csr")[0]+".crt"
+    upload_file_to_server(csr_filename, server_ip, username)
+
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as client_socket:
+        client_socket.connect((server_ip, CUSTOM_PORT))
+        message = f"CSR uploaded {csr_filename}".encode()
+        client_socket.send(message)
```
joenpera, line 83: send() - don't guarantee complete packet send.

https://docs.python.org/3/howto/sockets.html#using-a-socket

You need to have known message sizes. Use sendall() to send complete packet and on receiver side you need to know how long packet is coming.

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/secure_channel/secchannel.py
```diff
@@ -0,0 +1,74 @@
...
+        if not self._is_ssl_socket():
+            self.logger.error("Socket is not SSL enabled.")
+            return
+
+        if not self._is_socket_active():
+            self.logger.error("Socket is not active.")
+            return
+
+        try:
+            self.socket.send(message.encode())
```
joenpera, line 44: send() - don't guarantee complete packet send.

https://docs.python.org/3/howto/sockets.html#using-a-socket

You need to have known message sizes. Use sendall() to send complete packet and on receiver side you need to know how long packet is coming.

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/tools/wpactrl.py
```diff
@@ -0,0 +1,90 @@
...
+        self.socket.bind(self.local_path)
+        self.socket.connect(self.ctrl_path)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        os.unlink(self.local_path)
+        self.socket.close()
+
+    def request(self, cmd, msg_cb=None, reply_len=WPA_CTRL_MAX_REPLY_LEN):
+        self.socket.send(cmd)
```
joenpera, line 27: send() - don't guarantee complete packet send.

https://docs.python.org/3/howto/sockets.html#using-a-socket

You need to have known message sizes. Use sendall() to send complete packet and on receiver side you need to know how long packet is coming.

joenpera (changes requested):
send() - don't guarantee complete packet send.

https://docs.python.org/3/howto/sockets.html#using-a-socket

You need to have known message sizes. Use sendall() to send complete packet and on receiver side you need to know how long packet is coming.

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
TIISR: Great catch! But it shouldn't be possible to transmit data anyway if TLS verification failed, I imagine the other side will simply get a SSLError exception 🙂

joenpera: Doesn't justify wrong usage of socket.send()

TIISR: Sorry I didn't express myself very well, I meant there shouldn't be a send at all 😅

joenpera: --> remove socket.send() from here.

martin-tii (approved):

PR #338, opened 2023-11-17
Updated Jamming Avoidance code

dania-tii:
- Updated code to change channel via comms controller
- Updated osf startup function to remove manual startup (OSF network now mounts automatically if nRF52 firmware is flashed)
- Updated log file path to "/var/log"
- The jamming feature will start as a service - I will add the scripts to start the service in a separate pr


martin-tii (approved):

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/options.py
```diff
@@ -1,85 +1,24 @@
...
 from typing import List
-from log_config import logger
 
 VALID_CHANNELS = [36, 40, 44, 48, 149, 153, 157, 161]
 
-
 class Options:
     def __init__(self):
-        self.jamming_osf_orchestrator: str = 'fd01::1'
+        self.jamming_osf_orchestrator: str = 'fd04::1'
```
TIISR, line 7: OSF addressing has changed so this default is probably not correct anymore

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/jamming_setup.py
```diff
@@ -163,34 +171,129 @@ def start_jamming_scripts(args, osf_ipv6_address) -> None:
...
+    """
+    # Store the current working directory
+    current_directory = os.getcwd()
+
+    try:
+        # Change the directory to the scripts_path
+        os.chdir(args.nats_scripts_path)
+
+        # Define the commands to execute
+        command = ["python", "_cli_command_get_identity.py"]
```
joenpera, line 187: These test scripts can change and not meant to be used like this.

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/jamming_setup.py
```diff
@@ -21,9 +25,9 @@ def is_interface_up(interface) -> bool:
     return interface in available_interfaces
 
 
-def switch_frequency(args) -> None:
+async def switch_frequency(args) -> None:
```
joenpera, line 28: Is jamming avoidance running on top of upper macsec?

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/util.py
```diff
@@ -107,49 +113,26 @@ def map_freq_to_channel(freq: int) -> int:
...
-    cmd_iw_dev = ["iw", "dev"]
-    run_command(cmd_iw_dev, 'Failed to run iw dev')
+    try:
+        await nc.request(f"comms.channel_change.{config.MODULE_IDENTITY}", cmd.encode(), timeout=10)
+        logger.info(f"Published to comms.channel_change: {cmd}")
+        await asyncio.sleep(15)
+    except Exception as e:
+        logger.error(f"Failed to change channel: {e}")
+    finally:
+        await nc.close()
```
joenpera, line 135: There is a possibility also that TLS certificates are used for the communication.

example of TLS certificate usage:
https://github.com/tiiuae/mesh_com/blob/develop/modules/sc-mesh-secure-deployment/src/nats/scripts/client.py

Support both options:
- certificates available --> use TLS
- certificates not available --> don't use TLS

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/jamming_setup.py
```diff
@@ -163,34 +171,129 @@ def start_jamming_scripts(args, osf_ipv6_address) -> None:
...
+        logger.info("5 GHz channels must be of the following: (36,40,44,48,149,153,157,161)")
+        valid = False
+
+    # Return validity after all checks performed
+    return valid
+
+
+async def main():
     # Create Options instance
     args = Options()
```
joenpera, line 259: What are the arguments given to this scripts?   It is difficult to follow the code as there are args.<magic word> paramters.

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/jamming_setup.py
joenpera: Slack comments got from @TIISR to this. This interface is available also in new design.

TIISR: Yes and no: JA cannot be enabled by default, it needs to be enabled by user choice in MDM so it is a "red tier" feature in that sense, but JA won't be "running on top of upper macsec" in the sense of inspecting red network interfaces, it is meant to inspect black (or white in case we add something like HaLow there, as opposed to bat0 being the only white interface in most cases) network interfaces 🙂

TIISR: Sorry we were writing at the same time: my "yes and no" was in response to the top comment 🙂

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/jamming_setup.py
dania-tii: Ok, does this apply to client.py and config.py in nats/scripts as well?

joenpera: Yes,  nats/scripts are for testings the interface and those are same-time examples how to use the interface.

If someone decides to use the interfaces in there own code, similar functionality should done to own module as those testing tools can change (output which those are printing or even filenames can change).

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/options.py
dania-tii: Thanks for pointing this out, this argument should be configured for each specific device, so the current set value was for the particular device I was working on, I'll look into the new addressing

modules/sc-mesh-secure-deployment/src/2_0/features/jamming/jamming_setup.py
dania-tii: Currently there are no arguments given to this script, the variables should be edited before running the jamming avoidance based on the specific use-case

dania-tii:
Commit updates:

- updated osf address to reflect new addressing scheme
- updated test files
- added jamming_setup.py sigterm handler
- fine-tuning model for active scans

martin-tii (approved):

joenpera (approved):

PR #350, opened 2023-12-05
Fix comms settings bridge

joenpera:


joenpera:
More changes to remove double/triple quotes problem.

dania-tii (approved):

PR #351, opened 2023-12-10
Pypcap package installation

dania-tii:
Installing Pypcap package

- This package is required for LDPI (IDS)
- To install pypcap and its dependencies via buildroot python 2 needs to be installed, this will cause overhead/ performance issues. Instead, pypcap will be installed post-image build using its precompiled binary.


martin-tii (approved):

joenpera:
This is again only for aarch64.

joenpera (approved):

PR #358, opened 2023-12-14
Install scipy and scikit-learn packages

dania-tii:
Installing scipy and scikit-learn packages

- Scipy package is required for LDPI (IDS) and PHY-CRA/ RSS
- Scikit-learn is required for LDPI
- The scipy package was not successfully installing via buildroot, instead it will be installed upon device boot-up (scikit-learn depends on scipy so it will be installed upon-device boot up as well). I will continue trying to integrate scipy and scikit-learn via buildroot but until then this approach will be used.
- This installation is for aarch64, similarly these packages need to integrated for riscv.

### DIFF
--- a/modules/utils/package/install_dependencies.sh
+++ b/modules/utils/package/install_dependencies.sh
@@ -1,21 +1,34 @@
-# Install ml_packages
-cd /opt/mesh_com/modules/utils/package || exit
+# Install torch
 chmod +x /opt/mesh_com/modules/utils/package/ml_packages.tar.gz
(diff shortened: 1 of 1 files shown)


joenpera (approved):

martin-tii (approved):

dania-tii:
> Is the dependencies documented somewhere? Which features are dependent from these script installations?

At the moment jamming avoidance, PHY-CRA/ RSS and LDPI (IDS) all depend on packages from this script installation. The same script installation should be created for riscv. 

Do you suggest that we document the dependencies for all MS2.0 features in general or specifically the features that depend on the packages from this script installation?

joenpera:
> features that depend on the packages from this script installation

This one.

PR #361, opened 2023-12-15
Certificate bundle upload to server

joenpera:
- certificate bundle upload
- features json to yaml support
- dead code removal
- styling changes and imports optimized

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
@@ -9,6 +9,9 @@ import signal
 import ssl
 import subprocess
 import threading
+import os
+import glob
+import base64
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
 from typing import Optional
@@ -32,6 +35,7 @@ from src import comms_status
 #     "ENABLE_VISUALISATION",
 #     "DISABLE_VISUALISATION",
 # ]
+
 
 class MeshTelemetry:
     """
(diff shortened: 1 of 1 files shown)


modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
```diff
@@ -500,26 +499,35 @@ def __write_config_to_file(response: requests.Response) -> None:
...
-            # so bundling just device cert and ca cert for now.
-            with open(self.__certificate_file, 'r') as cert_file:
-                certificate_content = cert_file.read()
+            certificates_dict = {}
+
+            files = glob.glob("/opt/at_birth.tar.bz2*")
 
-            with open(self.__ca, 'r') as ca_file:
-                ca_certificate_content = ca_file.read()
+            if 4 > len(files) or len(files) == 0:
```
TIISR, line 515: Not urgent now but to avoid a risk of this maybe breaking later: we do not need exactly 4 files, we need the at_birth.tar.bz2 archive itself of course but with regards to the digital signatures we technically need only one, so it would be a pity to refuse to upload if some 3rd party software update left us with only 1 or 2 out of 3 signatures available :-)

TIISR (approved):

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
joenpera: I will remove the check.

joenpera: or change it to return fail if there is not any.

govindsi (approved):

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
joenpera: Fixed with  "if len(files) == 0  then return fail"

joenpera:
We need to wait Server side feature also to be available before merge.

andtokm (approved):

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
TIISR: Thank you :)

PR #362, opened 2023-12-15
Decision engine

SelinaTII:
This is the PR to merge the decision engine code and test scripts

- Currently, the decision engine is called with sample feature classes (for PHYCRA, RSS, IDS) from the MS2.0 main script: https://github.com/tiiuae/mesh_com/blob/decision_engine/modules/sc-mesh-secure-deployment/src/2_0/main.py
- Once this is merged, integration with actual feature classes will be done.


martin-tii (approved):

common/tests/functional/test_decision_engine/test_mba.py
saauvine, line 1: wlp1s0 interface is not available in CM2.0 and test is likely to fail with that HW.

SelinaTII: Updated the test code to use the first interface (among wlpXs0, halow1, eth0) available on the device instead of hard coding wlp1s0

saauvine (approved):

martin-tii (approved):

PR #364, opened 2023-12-18
Update OSF test script to check for and use latest IPv6 addressing scheme

mbaddeley:
Minor update to stop using the old fd0X addressing and start using the IPv6 address assigned to tun0.

### DIFF
--- a/common/tests/functional/test_osf.sh
+++ b/common/tests/functional/test_osf.sh
@@ -45,7 +45,7 @@ _test() {
 
   # Extract the inet6 IP address starting with 'fde3' from the 'tun0' interface
   ip_address=$(ifconfig tun0 | grep -o 'inet6 fde3:[^ ]*' | awk '{print $2}')
-  # Use ping6 to ping the IP address fd02::1 over the tun0 interface
+  # Use ping6 to ping the IP address over the tun0 interface
   if ! ping6 -c 3 -I tun0 "$ip_address" &>/dev/null; then
     echo "ping6 test to $ip_address over tun0 failed." | print_log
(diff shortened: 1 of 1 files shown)


Hiihtaja77 (commented):
# Use ping6 to ping the IP address fd02::1 over the tun0 interface
obsolete comment ?

Hardcoded prefix fde3:  . Tun0 can any other one from recommended list ( 15 variants)
Can it be done in any other way ?

mbaddeley:
> # Use ping6 to ping the IP address fd02::1 over the tun0 interface
> obsolete comment ?
> 
> Hardcoded prefix fde3: . Tun0 can any other one from recommended list ( 15 variants) Can it be done in any other way ?

I've removed the comment reference to fd02::1, but I think looking at the fde3 address is fine, right? This is agreed as the prefix for OSF.

Hiihtaja77:
In best case you should be able to extract this long address, modify it to node address and ping Node not Host.
but it can be 2 global addresses e.g 2001:db8: can be added for R&D
 May be have sense to extract first global address from tun0 what is not equal to 2001:db8
and use it to ping Host ( Node have other address)

Hiihtaja77 (commented):
ok

Hiihtaja77 (approved):

mbaddeley:
OK, redone the test to use slipcmd to get the node IP. Need to stop and start osf for this to stop slipcmd jamming tunslip6 (thanks @Hiihtaja77 for the tip). Tested with latest image as of 19/12/23 and passes.

ZAbidTii (approved):

PR #370, opened 2023-12-29
CBMA Unit and Functional Tests

pentestiing:
This PR moves the CBMA functional test folder to their proper place under `mesh_com/common/tests/functional`, unit tests ones under `mesh_com/modules/sc-mesh-secure-deployment/src/2_0/features/cbma/tests`, and also adds the `cbma_mdm_integration.sh` script to easily test that CBMA works with MDM :ice_cream:


common/tests/functional/test_cbma/cbma_mdm_integration.sh
```diff
@@ -0,0 +1,146 @@
...
+    echo "Usage: $0 <mdm-server-ip> <csl1-ip> [<csl2-ip> ...]" >&2
+    exit 1
+}
+
+setup_ssh() {
+    ip="$1"
+
+    echo -n "[+] Setting up SSH setup for ${ip}... "
+
+    command ssh "root@${ip}" "cat ~/.ssh/id_rsa" | ssh-add - >/dev/null 2>&1
```
TIISR, line 38: Not sure I understand this correctly: we are connecting to the remote host (first manually typing the password I guess) to cat their private key back into the local SSH agent? 🤔 I guess this technically works but only as long as the CSLs have that RSA key pair built in, which Govind promised is only in debug versions and which eventually needs to be replaced with certificates anyway... So for security reasons and especially if the remote .ssh/id_rsa file might not exist, what should be done instead is to add a local public key (the .ssh/id_*.pub files) to the remote .ssh/authorized_keys, which can be done with 'cat >>' of course but nowadays there is even the ssh-copy-id command to facilitate this 🙂

common/tests/functional/test_cbma/cbma_mdm_integration.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/sh
+
+
+MDM_AGENT_DIR="/opt/mesh_com/modules/sc-mesh-secure-deployment/src/nats"
+MDM_AGENT_CERTS="${MDM_AGENT_DIR}/certs"
+
+MDM_SERVER_DIR="~/mdm-server"
```
TIISR, line 7: Maybe nothing wrong here but in general best to avoid using tilde within variable contents as they are not always expanded, I don't remember in which case, maybe if you get executed with /bin/sh instead of Bash for example; safer to use $HOME instead 🙂 But then again that could end up being empty in some rare cases, which here would be making you write at the root of the file system 🤔

TIISR (approved):

ZAbidTii (approved):

common/tests/functional/test_cbma/cbma_mdm_integration.sh
pentestiing: Done!

Using in the example "\$HOME" instead of "$HOME" as it should expand in the SSH server :label:

common/tests/functional/test_cbma/cbma_mdm_integration.sh
pentestiing: Done!

It is missing the cleanup of the added key in `.ssh/authorized_keys`, maybe is not a big deal but let me know if you have concerns with having leftover keys there :dagger:

TIISR (approved):
Should delete the ephemeral pubkey from each destination .ssh/authorized_keys as opposed to adding them endlessly, but that's a clean-up for when we have time :)

PR #375, opened 2024-01-05
Updated paths to certificates used to sign and verify MBA messages as per the birth certificate location

SelinaTII:
Also updated the MBA tests accordingly


martin-tii (approved):

pentestiing (approved):
Awesome! :1st_place_medal:

PR #377, opened 2024-01-10
CBMA MDM Multiprocessing

pentestiing:
Behold fellow reader, the promised PR!

This PR includes some of the old problematic changes (now fixed) discussed in #337 alongside the multiprocessing additions in the old way of calling CBMA and the fixes implemented along the way (all of this squashed in the oldest commit for brevity but still accessible under https://github.com/tiiuae/mesh_com/tree/cbma-mdm-integration/). Another PR will follow this one with the error handling functionality.

Then, this PR also includes individual commits concerning the MDM agent side, which are more relevant for the current state of development.

Have fun! :star2:

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
+++ b/modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
@@ -7,7 +7,6 @@ import sys
 sys.path.insert(0, '../')
 from tools.verification_tools import *
 from tools.custom_logger import CustomLogger
-from tools.utils import wait_for_interface_to_be_pingable
 import glob
 
 logger_instance = CustomLogger("Server")
--- a/modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
+++ b/modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
@@ -15,6 +15,8 @@ logger = logger_instance.get_logger()
 
 
 class AuthServer:
+    CLIENT_TIMEOUT = 60
+
     def __init__(self, interface, ip_address, port, cert_path, ca_path, mua):
         threading.Thread.__init__(self)
         self.running = True
@@ -25,9 +27,11 @@ class AuthServer:
         self.interface = interface
         self.mymac = get_mac_addr(self.interface)
         # Create the SSL context here and set it as an instance variable
-        self.context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
+        self.context = ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH,
+                                                  cafile=glob.glob(self.ca)[0])
+        self.context.minimum_version = ssl.TLSVersion.TLSv1_3
+        self.context.check_hostname = False
         self.context.verify_mode = ssl.CERT_REQUIRED
-        self.context.load_verify_locations(glob.glob(self.ca)[0])
         self.context.load_cert_chain(
             certfile=glob.glob(f"{self.CERT_PATH}/MAC/{self.mymac}.crt")[0],
             keyfile=glob.glob(f"{self.CERT_PATH}/private.key")[0],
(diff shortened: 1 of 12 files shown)


modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
```diff
@@ -51,14 +55,14 @@ def handle_client(self, client_connection, client_address):
 
     def authenticate_client(self, client_connection, client_address, client_mac):
         secure_client_socket = self.context.wrap_socket(client_connection, server_side=True)
+        secure_client_socket.settimeout(self.CLIENT_TIMEOUT)
```
andtokm, line 58: CLIENT* named variable may looks a little misleading here, withing the Server class )

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/tools/utils.py
```diff
@@ -19,6 +20,78 @@
 logger = logger_instance.get_logger()
 
 
+######### MOVED FROM CLASSES #############
+
+def batman(batman_interface):
+    try:
+        subprocess.run(["ip", "link", "add", "name", batman_interface, "type", "batadv"], check=True)
+
+        batman_exec(batman_interface,"batman-adv")
```
andtokm, line 29: just to make sure: **batman_exec()** function sink's all _subprocess.CalledProcessError_ exceptions. 
It will just log some stuff but will not allow any _CalledProcessError_ leave the function )

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/auth/authServer.py
pentestiing: Done :+1:

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/tools/utils.py
pentestiing: I removed this unneeded try-except as this function is only being called by another one. Both can be joined (and maybe some more) but there are unit tests for these functions so I'll leave them be for now until I have time to do a proper refactor :+1:

saauvine:
Why are using 60s long socket timeouts? I'm worried does it cause that cbma processes stays alive that long time after trying to terminate them because long timeouts in blocking calls are affecting how frequently shutdown_event checks are performed. 

I tried to test that theory and run into problems when I tried to stop MDM agent after it was started with CBMA feature on. The problem I saw was that all the processes/threads were not stopped. Log indicates problem in /multicast.py", line 34 "OSError: no interface with this name". I did monitor processes/threads using htop via SSH connection (using usb0).

I suppose that this OS error is a consequence of running __cleanup_cbma method on MDM agent side right after trying to terminate cbma processes. => I tried to use process.kill() instead of process.terminate() in MDM agent's stop_cbma() method and it seemed to help for that particular problem but I think could be better to use shorter timeouts instead.

modules/sc-mesh-secure-deployment/src/2_0/features/cbma/tools/utils.py
andtokm: thx a lot )

andtokm (approved):

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
```diff
@@ -272,7 +276,7 @@ def __init__(
 
         self.__cbm_certs_path = "/opt/crypto/ecdsa/birth/filebased"
         self.__cbma_ca_cert_path = "/opt/mspki/ecdsa/certificate_chain.crt"
-        self.__cbma_threads: Dict[str, str] = {}
+        self.__cbma_processes: Dict[str, Process] = {}
         self.__cbma_shutdown_event = threading.Event()
```
saauvine, line 280: Should MDM agent should use multiprocessing.Event() instead since CBMA is running in different process. Not sure would it then require further changes on CBMA side to signal its threads.

saauvine (approved):

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
pentestiing: I removed it and changed the CBMA interface. This event is still useful inside CBMA (in changes found in another branch to be added in the future) for handling the threads on different error cases :cheese:

pentestiing:
> Why are using 60s long socket timeouts? I'm worried does it cause that cbma processes stays alive that long time after trying to terminate them because long timeouts in blocking calls are affecting how frequently shutdown_event checks are performed.
> 
> I tried to test that theory and run into problems when I tried to stop MDM agent after it was started with CBMA feature on. The problem I saw was that all the processes/threads were not stopped. Log indicates problem in /multicast.py", line 34 "OSError: no interface with this name". I did monitor processes/threads using htop via SSH connection (using usb0).
> 
> I suppose that this OS error is a consequence of running __cleanup_cbma method on MDM agent side right after trying to terminate cbma processes. => I tried to use process.kill() instead of process.terminate() in MDM agent's stop_cbma() method and it seemed to help for that particular problem but I think could be better to use shorter timeouts instead.

I made the tests once again and for now we cannot go less than 60 seconds for the client and server, otherwise the secure channel for the upper CBMA macsec key exchange is closed and the bridge never established. You can reduce it and it might work sometimes but not consistently. This will be fixed once the changes I mentioned in a comment above are added to CBMA :sandwich:

saauvine (approved):

PR #380, opened 2024-01-17
Use downloaded upper CBMA certificates

saauvine:
Jira-ID: SECO-3112, SECO-3111

"# FIXME: temporary hard coding for key path" indicates that private key handling will definitely change in the future. I didn't see it reasonable at this point to provide that from from MDM agent as handling of that will anyways change when key is used from HSM.

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
@@ -444,8 +444,7 @@ class MdmAgent:
                     await self.__loop_run_executor(
                         self.executor, ConfigType.DEBUG_CONFIG
                     )
-            # if self.__is_cbma_feature_enabled() and not self.__cbma_set_up:
-            if not self.__cbma_set_up:
+            if self.__is_cbma_feature_enabled() and not self.__cbma_set_up:
                 self.setup_cbma()
             await asyncio.sleep(
                 float(min(self.__interval, self.__debug_config_interval))
--- a/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
@@ -9,7 +9,9 @@ if __name__ == "__main__" and __package__ is None:
     sys.path.append(MS20_FEATURES_PATH)
     # Make it possible to run this module also in other folder (e.g. /opt/nats):
     sys.path.append("/opt/mesh_com/modules/sc-mesh-secure-deployment/src/2_0/features")
-    sys.path.append("/opt/mesh_com/modules/sc-mesh-secure-deployment/src/2_0/features/cbma")
+    sys.path.append(
+        "/opt/mesh_com/modules/sc-mesh-secure-deployment/src/2_0/features/cbma"
+    )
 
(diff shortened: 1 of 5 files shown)


TIISR (approved):
Thank you for specifying # FIXME: temporary hard coding for key path :)

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
```diff
@@ -308,26 +323,27 @@ def interface_monitor_cb(self, interfaces) -> None:
...
-        if self.__if_to_monitor is None:
-            self.start_service_discovery()
-        else:
-            for interface in self.__interfaces:
-                if interface.interface_name == self.__if_to_monitor:
-                    if interface.operstat == "UP":
-                        self.start_service_discovery()
-                    elif interface.operstat == "DOWN":
-                        self.stop_service_discovery()
+            if self.__if_to_monitor is None:
```
joenpera, line 338: Can we already here check is the discovery running and not to call start if the discovery is running?
Just causing extra function call... minor thing.

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
```diff
@@ -403,7 +419,13 @@ async def execute(self) -> None:
...
                             StatusType.DOWNLOAD_CERTIFICATES.value
                         ] = self.__action_certificates(resp, certificate_type.value)
-
+                        if (
+                            self.__status[StatusType.DOWNLOAD_CERTIFICATES.value]
+                            == "OK"
+                        ):
+                            # Restart CBMA with new certificates
+                            self.stop_cbma()
+                            self.setup_cbma()
```
pentestiing, line 428: Maybe this line can be removed now that `self.__cbma_set_up` set to False inside the `stop_cbma` function so that the `if` below starts CBMA again :raccoon:

pentestiing (approved):
Incredible! :dango:

modules/sc-mesh-secure-deployment/src/nats/src/comms_if_monitor.py
```diff
@@ -53,48 +55,54 @@ def monitor_interfaces(self):
...
                             existing_interface = next(
-                                (iface for iface in self.__interfaces if iface['interface_name'] == interface_info['interface_name']),
-                                None
+                                (
+                                    iface
+                                    for iface in self.__interfaces
+                                    if iface["interface_name"]
+                                    == interface_info["interface_name"]
+                                ),
+                                None,
```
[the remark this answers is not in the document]
saauvine: Admitted that previous version looks nicer. We've started to use black tool (with default line lenght) to do auto formatting and in this case it produced that looking code.

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
saauvine: My original intention was to leave setup_cbma() call to happen in that if clause later. This call is here because MDM server doesn't provide any kind of CBMA feature flag setting which causes that after connecting with MDM server we gent some configs from it that causes that  __is_cbma_feature_enabled() returns false thus CBMA would not be enabled. 

So yes, this part will be restructured in the future.

PR #382, opened 2024-01-18
Add exception catch to asyncio loop

joenpera:
Catching now possible exceptions.

Jira-Id: SECO-3346


saauvine (approved):

pentestiing (approved):
Bug-hunter! :bug: :star:

PR #384, opened 2024-01-19
SECO-3123: nrc1.5.0 update - setting long guard interval and support_ch_width

murugesan-kumar:
Due to latest halow nrc driver update, some of the default values are changed. To make it same as previous version, we need to apply these settings.


DamienDeMayaTII (approved):

joenpera (changes requested):
Is this ok for build targets using the 1.4 release?

murugesan-kumar:
> Is this ok for build targets using the 1.4 release?
Yes, this is ok for 1.4.0 release also.

joenpera (approved):

PR #389, opened 2024-01-23
Add validation for the delivered configuration message

joenpera:
- part of the validation is done in comms controller settings code, but empty message check was added to MDM agent side

Jira-Id: SECO-3293

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
@@ -1157,16 +1157,14 @@ class MdmAgent:
             del self.__cbma_processes[process_name]
 
     def __validate_response(
-        self, response: requests.Response, config: ConfigType, status_type: StatusType
+        self, response: requests.Response, config: ConfigType
     ) -> str:
         """
         Validate response
         :param response: HTTP response
         :param config: config
-        :param status_type: status type
         """
         # validation for response parameters
-
         self.__comms_controller.logger.debug(
             "validating response: %s, Config %s", response.text.strip(), config
         )
@@ -1197,6 +1195,8 @@ class MdmAgent:
                     self.__comms_controller.logger.error(
                         "Debug config field not found in config"
                     )
+            else:
+                self.__comms_controller.logger.error("Validation not implemented, unknown config")
         else:
             status = "FAIL"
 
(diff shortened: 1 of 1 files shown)


modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
```diff
@@ -1155,10 +1156,57 @@ def __cbma_process_alive_check(self):
...
+                    )
+            elif config == ConfigType.MESH_CONFIG:
+                try:
+                    if json.loads(response.text)["payload"]["radios"]:
+                        status = "OK"
+                except KeyError:
+                    self.__comms_controller.logger.error(
+                        "Radios field not found in config"
+                    )
+            elif config == ConfigType.DEBUG_CONFIG:
```
pentestiing, line 1192: Maybe for completeness and robustness add an `else` after this `elif` that fails with 'unknown config' or something catastrophically beautiful :fairy:

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
```diff
@@ -1155,10 +1156,57 @@ def __cbma_process_alive_check(self):
...
+        :param response: HTTP response
+        :param config: config
+        :param status_type: status type
+        """
+        # validation for response parameters
+
+        self.__comms_controller.logger.debug(
+            "validating response: %s, Config %s", response.text.strip(), config
+        )
+        status: str = "FAIL"
```
pentestiing, line 1173: `status_type` doesn't seem to be used in this function, is this just temporary or did you intend to pass it by reference (a workaround is to wrap it in a list) and change its value instead of using this local `status` variable? :thinking:

joenpera: I was first directly updating status from validation, but after removed it added return value.   Need to cleat that out.

joenpera: [34fcb4a](https://github.com/tiiuae/mesh_com/pull/389/commits/34fcb4a106727a4f90a0c8bc7a3291406f16aea1)

modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
joenpera: [34fcb4a](https://github.com/tiiuae/mesh_com/pull/389/commits/34fcb4a106727a4f90a0c8bc7a3291406f16aea1)

saauvine (approved):

pentestiing (approved):
:deciduous_tree:

PR #393, opened 2024-01-24
Fix correct MTU size for batman interface

joenpera:
Jira-Id: SECO-3431


pentestiing (approved):
Simple and delicious! :croissant:

PR #395, opened 2024-01-29
MDM/FMO/CBMA split and refactoring

joenpera:
Jira-Id: SECO-3162, SECO-3290


saauvine (approved):
Approving Mika's changes

PR #396, opened 2024-01-29
Fix problem where nodes were not on same channel after start

joenpera:
Jira-Id: SECO-3386


pentestiing (approved):
simplicity is delicious! :cookie:

TIISR (approved):
Great catch! 📶

PR #402, opened 2024-02-05
MS2.0 MDM/CBMA MVP Content

joenpera:
Contains changes related to tickets: 
SECO-3573, SECO-3578, SECO-3574, SECO-3578, SECO-3806, SECO-3497, SECO-3577, SECO-3802

Relates to tc_distro side PR 195

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
@@ -57,7 +57,15 @@ class CBMAControl:
 
         self.__white_interfaces = ["bat0", "halow1"]
         self.__red_interfaces = ["bat1", "wlan1"]
-        self.__na_cbma_interfaces = ["bat0", "bat1", "br-lan"]
+        self.__na_cbma_interfaces = [
+            "bat0",
+            "bat1",
+            "br-lan",
+            "eth0",
+            "eth1",
+            "usb0",
+            "lan1",
+        ]
 
         self.br_name = "br-lan"
         self.lower_batman = "bat0"
@@ -89,9 +97,9 @@ class CBMAControl:
             self.__interfaces.append(interface)
 
         self.logger.debug(
-                "__get_interfaces: %s",
-                interfaces,
-            )
+            "__get_interfaces: %s",
+            interfaces,
+        )
 
     def __create_bridge(self, bridge_name):
         ip = IPRoute()
--- a/modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
@@ -56,7 +56,7 @@ class CBMAControl:
         self.__upper_cbma_interfaces: List[Interface] = []
 
         self.__white_interfaces = ["bat0", "halow1", "wlp3s0"]
-        self.__red_interfaces = ["bat1", "eth0", "eth1", "usb0"]
+        self.__red_interfaces = ["bat1", "eth0", "eth1", "usb0", "lan1"]
         self.__na_cbma_interfaces = ["bat0", "bat1", "br-lan"]
 
         self.br_name = "br-lan"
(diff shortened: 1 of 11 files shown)


saauvine:
How shall we proceed with this PR? Review as is or should we put cbma still under that feature flag and the consider this as ready for review?

joenpera:
> How shall we proceed with this PR? Review as is or should we put cbma still under that feature flag and the consider this as ready for review?

If CBMA is kept disabled, then we need to change also default configuration generation to support CBMA disabled mode.  Now we are clearing some configuration items by assumption that CBMA is enabled.

saauvine:
> > How shall we proceed with this PR? Review as is or should we put cbma still under that feature flag and the consider this as ready for review?
> 
> If CBMA is kept disabled, then we need to change also default configuration generation to support CBMA disabled mode. Now we are clearing some configuration items by assumption that CBMA is enabled.

True and we probably don't want to put any energy to that. I hope nobody is developing things upon mesh-11s._nats.sh before this gets merged. Nobody is testing these changes before things gets merged.

pentestiing:
This PR won't be tested from our side until CBMA MVP is merged into develop. So, if you need some commits from here for any MDM MVP PRs you are working on, I'd recommend to cherry pick it and then remove it from this one :moon_cake:

joenpera:
> This PR won't be tested from our side until CBMA MVP is merged into develop. So, if you need some commits from here for any MDM MVP PRs you are working on, I'd recommend to cherry pick it and then remove it from this one 🥮

We are aware of the situation and we can wait this one and then do a big bang :boom:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
```diff
@@ -337,15 +341,15 @@ def __shutdown_and_delete_bridge(self, bridge_name):
...
-        self.__shutdown_interface(self.upper_batman)
+        self.__shutdown_interface(self.LOWER_BATMAN)
+        self.__shutdown_interface(self.UPPER_BATMAN)
         self.__delete_ebtables_rules()
         self.__delete_macsec_links()
-        self.__destroy_batman_interface(self.lower_batman)
-        self.__destroy_batman_interface(self.upper_batman)
+        self.__destroy_batman_interface(self.LOWER_BATMAN)
+        self.__destroy_batman_interface(self.UPPER_BATMAN)
         # HACK: How can we know that "br-upper" even exists and who owns it?
```
TIISR, line 350: It shouldn't: IIRC it was the name of the MACsec bridge in the Upper CBMA, but as part of the rewrite this is being renamed to the conflict-avoiding Xmb002091abcdef format where X is 'l' in Lower and 'u' in Upper, and 00:20:91:ab:cd:ef is the MAC address of the underlying interface. This is also for coherence with the other interfaces created as part of the "mess" between black & white or between white & red: MACVLAN interfaces prefixed with Xmv and MACsec interfaces prefixed with Xms (except Xmv and Xms names contain the _peer's_ MAC, so on a node called A under the umb002091aaaaaa bridge for example you might see ums002091bbbbbb and umv002091bbbbbb for a B peer, um[sv]002091cccccc for a C peer, and so on). The mess of [lu]m[bsv]* interfaces is handled within CBMA, CommsController only tells CBMA something like "you're in charge of between wlan0 and bat0", another "between bat0 and bat1", yet another "between halow0 and bat1" and so on.

TL;DR: please delete br-upper :)

TIISR (approved):

modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
saauvine: Deletion of br-upper has to be kept for now in cbma_controller's __cleanup_cbma method as current version of CBMA creates it and in case we restart MDM agent we need to ensure that everything CBMA related is cleared before trying to set it up again. I've mentioned it to be a hack as CBMA feature code is the one who has possibly created it (same applies to ebtables). Logically CBMA feature code should take care of cleaning what is has created when CBMA processes are terminated. Once CBMA rewrite code available for integration we can remove that.

saauvine (approved):

modules/utils/docker/entrypoint_nats.sh
```diff
@@ -33,10 +33,31 @@ else
       sleep 1
   done
 
+  CBMA=$(extract_features_value "CBMA" $YAML_FILE)
```
TIISR, line 36: Hopefully we never use spaces or other problematic characters in filenames but just in case, to be safe it should be "$YAML_FILE" as in between double quotes :-)

modules/utils/docker/entrypoint_nats.sh
```diff
@@ -33,10 +33,31 @@ else
...
+  if [ "$CBMA" == "false" ]; then
```
TIISR, line 37: Probably not important here since hopefully there won't be a need to disable CBMA, but for other such tests in the future: it might be worth making the test case insensitive, otherwise someone might write FEATURE=False or FEATURE=FALSE or maybe even FEATURE=0 and not realize that these do not disable the feature.

TIISR (approved):

modules/utils/docker/entrypoint_nats.sh
joenpera: CBMA setting check will be removed once CBMA is default config.

modules/utils/docker/entrypoint_nats.sh
joenpera: CBMA setting check will be removed once CBMA is default config.

modules/utils/docker/entrypoint_nats.sh
joenpera: SECO-4104 added to backlog.

PR #403, opened 2024-02-06
TEMP: remove eth0 from red bridge to avoid breaking PXE boot

pentestiing:
The title :upside_down_face: 

Basically this is also the place to discuss why do this or not or when or workarounds :man_shrugging:

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/src/cbma_controller.py
@@ -493,7 +493,7 @@ class CBMAControl:
 
         # FIXME: Hard coded red interfaces for testing purposes
         # Eventually MDM server will define interface colors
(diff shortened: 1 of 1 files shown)


TIISR (approved):
Just checked with Michael: no developer is using CM1.5, but even if someone is, they can actually get an additional RJ45 in CM1.5 simply by using one of the "industrial connector to RJ45" yellow cables that we have already at least in the cupboards of the Wireless Room on TII 1st floor :) 

The only good solution in general is to setup networking (bridge, CBMA and whatnot) from an Initial RAM Disk (initrd) that all Linuxes have had since forever but that unfortunately HardenedOS1 lacks.

saauvine (approved):

DamienDeMayaTII:
@pentestiing, still needed or discarded?

pentestiing:
Closing as a better way of doing it is added in https://github.com/tiiuae/mesh_com/pull/402 :abacus:

PR #405, opened 2024-02-08
utm

DaniilTroshkovTII:
UTM config support

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/comms_nats_controller.py
@@ -1412,14 +1412,16 @@ class MdmAgent:
                 self.logger.debug("config: %s, ret: %s", config, ret)
                 if ret == "OK":
                     self.__status[status_type] = "OK"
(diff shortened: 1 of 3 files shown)


DamienDeMayaTII:
@DaniilTroshkovTII, still in draft?

DaniilTroshkovTII:
> @DaniilTroshkovTII, still in draft?

Hello @DamienDeMayaTII I would prefer to merge this changes but I am not sure if @govindsi planning to merge it. In my view we can afford this small overhead of requesting MDM server for UTM config - for example if server do not support it (returns 405) MDM agent can just ignore it (make polling time s equal to OK polling time which is 10 mins). So this overhead request will only appears once in 10 minutes. From my view it will be easy to support this additional UTM functionality in MDM agent.

DamienDeMayaTII:
@DaniilTroshkovTII , please check with Seb and kindly decide to merge or cancel. Thanks!

DamienDeMayaTII:
@DaniilTroshkovTII , any update?

govindsi:
This will not land in develop as this is meant for staging feature branch. After MVP, we will only have lte only MDM config for UTM specific use case.

PR #408, opened 2024-02-19
Add test for stop/start of OSF driver

Hiihtaja77:
SECO-3740
configuration tests are renamed also.


mbaddeley (commented):
What's the difference between `1M_1.sh` and `1M_2.sh`?

Hiihtaja77:
BLE_1M in both cases, But the hopping channels set and keys are different. It is possible to configure 2 osf networks what are not interfere each other.

mbaddeley (approved):

PR #424, opened 2024-03-23
Upgrade CBMA to v0.1.2

pentestiing:
This PR upgrades CBMA to v0.1.2, some highlights:
- Improves network connectivity by temporarily stop sending multicast packets when in different phases
- Temporarily ignores packets from peers that fail to authenticate repeatedly
- Adds extra interfaces in the CBMAController for the MDM agent to query the status of a specific CBMA instance
- Improvements in the auxiliary dev scripts

And, as always, other fixes, refactoring and robustness additions here and there!

:call_me_hand:


TIISR (approved):

METII99 (approved):
Will debug later an inconsistency in upper cbma neighbours (maybe it is a non-issue , Let's C) but approving to make it to midnight build assuming it doesn't fail 🦀🐚

PR #425, opened 2024-03-26
The source file 

francislin12:
I have just changed the source of the configuration file


francislin12:
I have just added the source of the configuration file

joenpera:
Which interface you are planning to use for olsr?  This file is only starting black interface wifi and that will be used later for cbma.

francislin12:
@joenpera all interfaces

joenpera (changes requested):
This don't add any functionality as sourced functions are not called.

Prpbably proper OLSRController Class python implementation needed for olsr control.  Similar which was done for CBMA (CBMAController).   Then that functionality can be integrated to MDM side.

saauvine:
Closing as not valid

PR #429, opened 2024-03-29
Make MDM Agent wait until interfaces are ready before starting CBMA

pentestiing:
This PR fixes the issue where the MDM Agent starts CBMA before the interfaces are ready.

Jira-ID: SECO-4734

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
@@ -6,6 +6,7 @@ import logging
 import subprocess
 import threading
 import time
+import socket
 from typing import List, Union
 from copy import deepcopy
 import json
@@ -432,7 +433,7 @@ class CBMAAdaptation(object):
                     for interface in self.__interfaces
                 )
                 if found:
-                    return True
+                    break
             elapsed_time = time.time() - start_time
 
             if elapsed_time >= timeout:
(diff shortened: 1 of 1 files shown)


TIISR (approved):

pentestiing:
As I realize it might not be clear what the issue is and how this PR is solving it:
- Just after booting up, some interfaces might still not be ready for socket operations when the MDM Agent is started.
- As a result, when CBMA gets launched in them and try to perform socket operations (such as creating the multicast listener), the OS will raise an exception and the CBMA instance will gracefully stop
- The MDM Agent still doesn't have functionality to monitor the state of the interfaces and restart CBMA in them if needed, so this interface will remain without CBMA running on it

:shell:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
```diff
@@ -441,6 +442,27 @@ def __wait_for_interface(self, if_name, timeout: int = 3) -> bool:
...
+        index = socket.if_nametoindex(if_name)
+
+        # Now we wait until the interface is ready
+        start_time = time.time()
+        while True:
+            try:
+                socket.create_server((link_local_address, 0, 0, index),
+                                     family=socket.AF_INET6)
+                return True
+            except Exception:
```
andtokm, line 460: is it not too broad exception clause we have here ? ))

andtokm (approved):

PR #430, opened 2024-04-01
Fix missing index variable in MDM Agent Global IPv6 assignment

pentestiing:
This PR fixes a regression introduced in https://github.com/tiiuae/mesh_com/pull/429 that prevented the assignment of global IPv6 addresses on MDM agent startup :adhesive_bandage:


METII99 (approved):
🏃

joenpera (approved):

PR #435, opened 2024-04-05
Fix MDM Agent process stop

joenpera:
Jira-Id: SECO-4737

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
@@ -831,8 +831,6 @@ async def main_mdm(keyfile=None, certfile=None, ca_file=None, interface=None) ->
     """
     cc = comms_controller.CommsController()
 
-    mdm_agent_stopped = False
-
     if keyfile is None or certfile is None or ca_file is None:
         cc.logger.debug("MDM: Closing as no certificates provided")
         return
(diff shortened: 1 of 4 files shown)


modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
```diff
@@ -854,24 +841,36 @@ async def stop():
...
 
-    def signal_handler():
+
+    def signal_handler(signum, frame):
+        global mdm_agent_stopped
+
+        sig_name = signal.Signals(signum).name
+        cc.logger.debug(f"signal_handler {sig_name} ({signum}) received.")
+
+        if sig_name == "SIGUSR1":
```
pentestiing, line 851: Shouldn't SIGINT and SIGTERM also be checked for here? Maybe someone gets impatient and tries to stop the MDM agent twice in a row :thinking:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
```diff
@@ -540,7 +558,7 @@ def __setup_radios(self) -> bool:
 
         return True
 
-    def setup_cbma(self) -> None:
+    def setup_cbma(self) -> bool:
```
pentestiing, line 561: It doesn't look like this function can return `False` :heavy_check_mark:

modules/sc-mesh-secure-deployment/src/nats/src/comms_if_monitor.py
```diff
@@ -94,6 +96,24 @@ def monitor_interfaces(self):
...
+    def __create_dummy_interface(self, interface_name):
+        """
+        Create a dummy interface.
+
+        Arguments:
+            interface_name: str -- Name of the interface to be created.
+        """
+        subprocess.run(['ip', 'link', 'add', 'dev', interface_name, 'type', 'dummy'],
+                           check=True)
+    def __delete_dummy_interface(self, interface_name):
```
pentestiing, line 108: Newline missing before :baby_bottle:

pentestiing (approved):
Supremely superb! :superhero:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
joenpera: yes, but there is "return true" at the end and code editors are complaining "-> None vs return true".   Fail case handling needs to added.

saauvine (approved):

PR #441, opened 2024-04-22
Upgrade CBMA to v0.1.3: Add MACsec bridge to batman on startup

pentestiing:
This PR makes CBMA early-create the MACsec bridge derived from the underlying CBMA interface!

As a result, there will be no more these misunderstandings about whether the batman interface exist/works or not when doing `batctl n` :bat: 

Jira-ID: SECO-4889


johan-juntunen (approved):

andtokm (approved):

modules/sc-mesh-secure-deployment/src/nats/cbma/scripts/mess/cleanup_bridge.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [ $# -ne 2 ]; then
+    >&2 echo "Usage: $0 u bat0"
+    >&2 echo "Usage: $0 l wlan0"
+    exit 1
+fi
+export L_OR_U="$1"
+export BASE_INTERFACE_NAME="$2"
+
+MAC_REGEX='[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}:[0-9a-f]{2}'
```
joenpera, line 26: Could it be:
MAC_REGEX='([[:xdigit:]]{2}:){5}[[:xdigit:]]{2}'

joenpera (commented):
LGTM, Didn't try it.

modules/sc-mesh-secure-deployment/src/nats/cbma/scripts/mess/cleanup_bridge.sh
pentestiing: I'm always forgetting about the regex hex character class, great reminder! :+1: 

The new bridge scripts are actually literal copy-paste + stripping from other "mess" scripts, so yeah, some refactoring would be nice in them. For now I will leave this task for when the MACsec key refresh is implemented properly :angel:

PR #442, opened 2024-04-24
Add support to configure hop penalty values via config file

saauvine:
Jira-IDs: SECO-4849, SECO-268, SECO-4947, SECO-4948

Batman related control moved from cbma_adaptation module to bat_ctrl_utils module. Added support define hop penalty settings via config file. Unit tests created for bat_ctrl_utils. ipv6 support added for vlan creation.


modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
```diff
@@ -0,0 +1,280 @@
...
+                    except Exception as e:
+                        self.logger.info(
+                            "Failed to set hop penalty %s for: %s. Error: %s",
+                            penalty,
+                            interface,
+                            e,
+                        )
+
+    @staticmethod
+    def __get_interface_mac(interface: str) -> Union[None, str]:
```
joenpera, line 198: Maybe not bat_ctrl related functionality.  Probably we need network_utils also for common networking methods.

joenpera: No need to fix it now.

joenpera (approved):

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
```diff
@@ -0,0 +1,280 @@
...
+        try:
+            with open(
+                file=f"/sys/class/net/{interface}/address",
+                mode="r",
+                encoding="utf-8",
+            ) as file:
+                return file.readline().strip()
+        except FileNotFoundError:
+            return None
```
pentestiing, line 207: I'd say that an exception should be raised if an interface doesn't have a MAC address as this should never be expected behavior, unless this function is called with potentially non-existing interfaces :paintbrush:

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
```diff
@@ -0,0 +1,280 @@
...
+        self.logger.info(f"hardif_hop_penalty: {hardif_hop_penalty}")
+        if hardif_hop_penalty is not None:
+            for interface, penalty in hardif_hop_penalty.items():
+                if penalty is not None:
+                    try:
+                        # Serch hardif from all my batmans
+                        for batman_interface in self.__bat_interfaces:
+                            hardif = self.__find_batman_hardif(interface, batman_interface)
+                            if hardif:
+                                subprocess.run(
```
pentestiing, line 178: I find this indentation a bit extreme, any chance you could early return if for example `hardif_hop_penalty` is `None`, etc, to avoid this? :dancing_women:

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
```diff
@@ -0,0 +1,280 @@
...
+        except subprocess.CalledProcessError as e:
+            self.logger.warning(
+                "Interface %s not found in %s interface list. Error: %s",
+                interface,
+                batman_if,
+                e,
+            )
+
+        # If interface is not found, try to find it using MAC address
+        mac_address = self.__get_interface_mac(interface).replace(":", "")
```
pentestiing, line 229: As `__get_interface_mac` can return `None`, you have to move the `replace` inside the `if` below to avoid an exception :package:

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
```diff
@@ -0,0 +1,280 @@
...
+                    f"Find hardif with mac: {mac_address} from {batman_if}"
+                )
+                batctl_output = subprocess.check_output(
+                    ["batctl", batman_if, "if"]
+                ).decode("utf-8")
+                # batctl_output can contain multpile lines with syntax like:
+                # lmb00301a4fc7d5: active
+                for line in batctl_output.splitlines():
+                    # Get characters before colon characters
+                    interface_match = re.search(r"^(.*?):", line)
```
pentestiing, line 242: It seems a bit of a beast to use a regex in this case as the output of `batctl <bat> if` is rather simplistic...
Can't you use `interface_name = line.split(":")[0]` or just have it like `if mac_address in line:`? :kangaroo:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
```diff
@@ -150,9 +146,7 @@ def __create_vlan_interfaces(self) -> None:
                     )
 
                     # Bring VLAN interface up
-                    subprocess.run(
-                        ["ip", "link", "set", vlan_name, "up"], check=True
-                    )
+                    subprocess.run(["ip", "link", "set", vlan_name, "up"], check=True)
```
pentestiing, line 149: I think this might make the PEP linter tool unhappy :cake:

modules/sc-mesh-secure-deployment/src/nats/src/constants.py
```diff
@@ -41,17 +43,35 @@ class Constants(Enum):
...
+    DOWNLOADED_RSA_CERTS_PATH: str = (
+        DOWNLOADED_CBMA_UPPER_PATH + "/crypto/rsa/birth/filebased"
+    )
+    DOWNLOADED_UPPER_CBMA_CA_CERT = (
+        DOWNLOADED_CBMA_UPPER_PATH + "/upper_certificates/rootCA.crt"
+    )
     GENERATED_CERTS_PATH: str = f"{ROOT_PATH}/crypto"
+    ECDSA_BIRTH_FILEBASED: str = "/opt/crypto/ecdsa/birth/filebased"
+    RSA_BIRTH_FILEBASED: str = "/opt/crypto/rsa/birth/filebased"
+    RSA_BIRTH_KEY = GENERATED_CERTS_PATH + "/rsa/birth/filebased"
```
pentestiing, line 61: Add the `:str` type for consistency :bricks:

pentestiing (changes requested):
:ice_cube:

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
saauvine: That function can be indeed called with non-existing interface names as it is used in __find_batman_hardif which is called as a result of found batman configurations that can be made for example for CM2 that has halow1 but code is rung with Sleeve with wlp1s0 only.

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
saauvine: Agreed

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
saauvine: Agreed

modules/sc-mesh-secure-deployment/src/nats/src/constants.py
saauvine: Thanks for spotting that. Will change.

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
saauvine: Will check

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
saauvine: True. Will change.

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
```diff
@@ -0,0 +1,280 @@
...
+                self.__batman_ra = batman_ra
+        # Set batman routing algo is set before any interfaces are created
+        self.__set_batman_routing_algo()
+
+    def create_batman_interface(
+        self, batman_if: str, mac_addr: Optional[str] = None
+    ) -> None:
+        """
+        Creates a Batman interface with given name and mac.
+        If mac is not given then one is genereted.
```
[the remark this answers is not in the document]
saauvine: Fixed via cc2d5663f8ca5c884be02417f1011a5697087166

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
saauvine: Changed as suggested via cc2d5663f8ca5c884be02417f1011a5697087166

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
saauvine: Fixed via cc2d5663f8ca5c884be02417f1011a5697087166

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
saauvine: Fixed via cc2d5663f8ca5c884be02417f1011a5697087166

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
saauvine: I don't follow. Or are you referring to line length? I've been using black tool with default setting that limits line length to 88 instead of 79 that is what flake8 uses by default. Pylint itself is not complaining.

modules/sc-mesh-secure-deployment/src/nats/src/constants.py
saauvine: Fixed via cc2d5663f8ca5c884be02417f1011a5697087166

pentestiing (approved):
Very romantic :rose:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
pentestiing: I was referring to the line length, I thought you were using 79, no problem then :pie:

modules/sc-mesh-secure-deployment/src/nats/src/bat_ctrl_utils.py
pentestiing: The `if` inside the loop could be negated and use `continue`, but I'd say it's ok like this :dango:

PR #447, opened 2024-04-30
Fix AP handling and remove ctrl_interface as not needed for ms2.0

joenpera:
Jira-Id: SECO-4998

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/initd/S90APoint
+++ b/modules/sc-mesh-secure-deployment/src/nats/initd/S90APoint
@@ -43,7 +43,7 @@ case "$1" in
                 ${user:+--user $user} ${chroot:+--chroot $chroot} \
                 ${chdir:+--chdir $chdir} ${umask:+--umask $umask} \
                 ${stdout:+--stdout $stdout} ${stderr:+--stderr $stderr} \
-                -- rm -fr /var/run/hostapd/wlan1 ; $DIR$DAEMON ${!_MODE} $RADIO_INDEX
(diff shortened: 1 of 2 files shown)


modules/sc-mesh-secure-deployment/src/nats/initd/S90APoint
```diff
@@ -53,21 +53,19 @@ case "$1" in
...
-
     stop)
         if "$daemon" --running --name "$name" --pidfiles "$pidfiles"
         then
             echo "Stopping $name..."
             "$daemon" --stop --name "$name" --pidfiles "$pidfiles"
             echo done.
 
-            NEW_PID=$(ps ax | grep -E "[h]ostapd-$RADIO_INDEX" | awk {'print $1'})
+            NEW_PID=$(ps ax | grep -E "[h]ostapd-$RADIO_INDEX" | awk "{'print $1'}")
```
pentestiing, line 63: The `awk` command will fail, replace it to: `awk '{print $1}'`

Or even join it with the `grep`: `ps ax | awk "/ostapd-$RADIO_INDEX/{print \$1}"`

:bagel:

pentestiing (changes requested):
:call_me_hand:

modules/sc-mesh-secure-deployment/src/nats/initd/S90APoint
joenpera: Yes, this PR is draft and under testing currently and not including the final changes.

saauvine (approved):

pentestiing (approved):
:signal_strength:

PR #450, opened 2024-05-01
Use hardcoded fallback when MDM Server mDNS resolution fails

pentestiing:
Jira-Id: SECO-5072

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
@@ -80,7 +80,7 @@ class MdmAgent:
             self.__lock
         )
         self.__url: str = "defaultmdm.local:5000"  # mDNS callback updates this one
-        self.__fallback_url: str = f"[{self.cbma_ctrl.__IPV6_WHITE_PREFIX}::1]:5000"
+        self.__fallback_url: str = f"[{Constants.IPV6_WHITE_PREFIX.value}::1]:5000"
 
         self.service_monitor = comms_service_discovery.CommsServiceMonitor(
             service_name="MDM Service",
--- a/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
@@ -65,7 +65,6 @@ class MdmAgent:
         self.__interval: int = Constants.FAIL_POLLING_TIME_SECONDS.value  # TODO: possible to update from MDM server?
         self.__debug_config_interval: int = Constants.FAIL_POLLING_TIME_SECONDS.value  # TODO: possible to update from MDM server?
 
-        self.__url: str = "defaultmdm.local:5000"  # mDNS callback updates this one
         self.__keyfile: str = keyfile
         self.__certificate_file: str = certificate_file
         self.__ca: str = ca_file
@@ -80,6 +79,8 @@ class MdmAgent:
             self.logger,
             self.__lock
         )
+        self.__url: str = "defaultmdm.local:5000"  # mDNS callback updates this one
+        self.__fallback_url: str = f"[{self.cbma_ctrl.__IPV6_WHITE_PREFIX}::1]:5000"
 
         self.service_monitor = comms_service_discovery.CommsServiceMonitor(
             service_name="MDM Service",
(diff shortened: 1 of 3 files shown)


TIISR (changes requested):
Lol sorry I say "fdbb" for short but it should be the same 64bit prefix fdbb:... and only then ::1 🙂

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
```diff
@@ -66,6 +66,7 @@ def __init__(
         self.__debug_config_interval: int = Constants.FAIL_POLLING_TIME_SECONDS.value  # TODO: possible to update from MDM server?
 
         self.__url: str = "defaultmdm.local:5000"  # mDNS callback updates this one
+        self.__fallback_url: str = "[fdbb::1]:5000"
```
saauvine, line 69: Not sure did I understand Seb's comment correctly but should this be then:
self.__fallback_url: str = f"[{self.__IPV6_WHITE_PREFIX}::1]:5000"

pentestiing: I think so, will clarify now :+1:

TIISR (approved):

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
saauvine, line 1: I didn't remember that __IPV6_WHITE_PREFIX was defined on CBMA adaptation side thus now getting warning on MDM agent side about accessing protected member. Perhaps we should move those constant definitions later into constants.py module.

saauvine (approved):

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
```diff
@@ -80,6 +79,8 @@ def __init__(
             self.logger,
             self.__lock
         )
+        self.__url: str = "defaultmdm.local:5000"  # mDNS callback updates this one
+        self.__fallback_url: str = f"[{self.cbma_ctrl.__IPV6_WHITE_PREFIX}::1]:5000"
```
joenpera, line 83: This line won't work.

joenpera (changes requested):

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
pentestiing: Is the warning still there if they get moved to class variables?

```python
class CBMAAdaptation(object):
    """
    CBMA Control
    """
    __IPV6_WHITE_PREFIX: str = "fdbb:1ef7:9d6f:e05d"
    __IPV6_RED_PREFIX: str = "fdd8:84dc:fe30:7bf2"
...
```
:snake:

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
pentestiing: Can you try this? https://github.com/tiiuae/mesh_com/pull/450#discussion_r1587258156

joenpera: it's still private.

saauvine: It should be self.__fallback_url: str = f"[{self.cbma_ctrl._CBMAAdaptation__IPV6_WHITE_PREFIX}::1]:5000" in order to work. But better way would be probably move those constant definitions to constants.py instead. Perhaps also some of the other magic prefixes from 99_at_boot could be given some purpose via naming some of those as well.

pentestiing: Fixed in https://github.com/tiiuae/mesh_com/pull/450/commits/f09732dc32dcdab1e314eaee41e5af4f41e5f59f :1st_place_medal:

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
pentestiing: Moved to Constants in https://github.com/tiiuae/mesh_com/pull/450/commits/f09732dc32dcdab1e314eaee41e5af4f41e5f59f :ladder:

saauvine (approved):

joenpera (approved):

PR #451, opened 2024-05-02
Change MDM agent default None startup parameter to bat0

joenpera:
Jira-Id: SECO-5052, SECO-5093


modules/sc-mesh-secure-deployment/src/nats/initd/S90mdm_agent
```diff
@@ -18,9 +18,9 @@ BC_CERT_FILE=$(find /opt/crypto/rsa/birth/filebased/DNS/ -type f -name '*.local.
 BC_CA_FILE="/opt/mspki/rsa/certificate_chain.crt"
 
 if [ -e "$BC_CA_FILE" ] && [ -e "$BC_KEY_FILE" ] && [ -e "$BC_CERT_FILE" ]; then
-  command_args="-k $BC_KEY_FILE -c $BC_CERT_FILE -r $BC_CA_FILE"
+  command_args="-k $BC_KEY_FILE -c $BC_CERT_FILE -r $BC_CA_FILE -i bat0"
```
saauvine, line 21: This will indeed cause that MDM agent tries to monitor MDM server via bat0 interface. However, there are other issues that effectively prevents it from functioning.  MDM agent starts sevice monitoring for the wanted interface when interface status is UP. This doesn't work currently with Batman interfaces as they seem to remain in UNKNOWN state. Yet another issue is that MDM agent tries to stop monitoring only when interface state becomes DOWN and that doesn't work with for example LAN1 when one detaches cable. I am already working on those other issues.

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
```diff
@@ -229,15 +232,24 @@ def interface_monitor_cb(self, interfaces) -> None:
...
+                                self.mdm_service_available = False
+                                self.stop_service_discovery()
             else:
-                for interface in self.__interfaces:
-                    if interface.interface_name == self.__if_to_monitor:
-                        if interface.operstat == "UP":
-                            self.start_service_discovery()
-                        elif interface.operstat == "DOWN":
-                            self.stop_service_discovery()
+                self.mdm_service_available = False
```
joenpera, line 251: needs threading.lock as same parameter changed from service callback.

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
```diff
@@ -209,6 +210,8 @@ def __get_server_cert_type(self):
                 self.logger.error(
                     "Error occurred during certificate type detection: %s", e
                 )
+            finally:
+                sock.close()
```
pentestiing, line 214: I'd say to wrap the `sock.close()` in another `try... except` because if the socket cannot be created due to "Too many open files" or whatever, then closing it will cause an exception :baby_chick:

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
```diff
@@ -247,8 +260,13 @@ def start_service_discovery(self):
         """
         if not self.service_monitor.running:
             self.logger.debug("Start service monitor thread")
-
-            self.__service_monitor_thread.start()
+            if not self.__service_monitor_thread.is_alive():
+                self.__service_monitor_thread = threading.Thread(
+                    target=self.service_monitor.run
+                )
```
pentestiing, line 266: I'd suggest using a ThreadPool or using a loop with a Queue instead of creating a new thread every time as it seems their memory is not released by Python when they stop :cactus:

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
```diff
@@ -880,12 +898,19 @@ def signal_handler(signum, frame):
     parser.add_argument("-k", "--keyfile", help="TLS keyfile", required=False)
     parser.add_argument("-c", "--certfile", help="TLS certfile", required=False)
     parser.add_argument("-r", "--ca", help="ca certfile", required=False)
-    parser.add_argument("-i", "--interface", help="e.g. br-lan or bat0", required=False)
+    parser.add_argument(
+        "-i",
+        "--interface",
+        help="Interface from where to monitor MDM server. Default is bat0",
```
pentestiing, line 904: You can add a `default=Constants.LOWER_BATMAN.value` instead of manually doing it in an `if` below :waffle:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
```diff
@@ -856,8 +856,6 @@ def stop_cbma(self) -> None:
         Stops CBMA by terminating CBMA processes. Also
         destroys batman and bridge interfaces.
         """
-        if not self.__cbma_set_up:
```
pentestiing, line 859: I believe `self.__cbma_set_up` is now unused in CBMAAdaptation, plz remove it if so :2nd_place_medal:

pentestiing (changes requested):
Lucky draw :game_die:

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
pentestiing: As tested by @saauvine, it seems that this issue no longer affects modern Python3 :snake:

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
saauvine: This was good point. Fixed in commit 1826923e3d821a8ab2c2b156ce5c2d2db02a0359.

modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
saauvine: Fixed in commit 1826923e3d821a8ab2c2b156ce5c2d2db02a0359.

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
saauvine: Removed in commit 1826923e3d821a8ab2c2b156ce5c2d2db02a0359.

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
```diff
@@ -646,10 +647,7 @@ def setup_cbma(self) -> bool:
         # Set batman hop penalty
         self.__batman.set_hop_penalty()
 
-        self.__cbma_set_up = True
-
-        # todo what about the false return value?
-        return True
+        return lower_cbma_set_up and upper_cbma_set_up
```
pentestiing, line 650: I believe we should `return True` here and early `return False` if setting up lower or upper-CBMA fails:
```python
if not self.__setup_lower_cbma() or not self.__setup_upper_cbma():
    return False
```
:trackball:

pentestiing (changes requested):
:garlic:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
saauvine: True, it would make more sense but to be honest error handling would still require more work. As I run into SECO-5165 I run into a problem that upper CBMA setup started to fail because new certificates are invalid. Now with this implementation it causes setup_cbma() to fail which again leads to a situation that MDM agent tries again after its fail timeout (1s) which then leads yet to another problems because for the new round we still have for example lmb interfaces created in previous lower cbma setup round which leads to a situation that lmb interfaces are treated as black interfaces thus they are tried to be added to lower cbma which causes problems and subseguent trial to setup upper cbma would still fail due to invalid certificates. I was thinking to improve those things whilst doing unit tests for CBMA Adaptation. However, I don't know yet how I should deal cases where received upper CBMA certificates are invalid. Should I juts try to revert to birth certs? I created a seprate ticket for further improvents. SECO-5182.

saauvine: setup_cbma() updated as Carlos suggest with one addition to call stop_cbma() in case of failure. Also stop_cbma() itself was refactored a bit to use separate helper method to stop cbma.

pentestiing: Hmmm, I believe the `lmb` interface should be deleted if the associated CBMA instance is stopped (unless there's something else attached to the bridge not added by CBMA when it stops), so the scenario you described shouldn't happen if all CBMA instances are stopped if this method returns `False` before trying again...
Alternatively, maybe the `lmb` interfaces should be added to the `__na_cbma_interfaces` list (and it shall support wildcards) :thinking:

pentestiing (approved):
:pig2:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
pentestiing: Ah, stopping all CBMA instances in case of failure is what you did in [48c4ea9](https://github.com/tiiuae/mesh_com/pull/451/commits/48c4ea964d9ab59e16213b7e62dc0d5983a798b0), awesome! :star2:

pentestiing (approved):
:face_in_clouds:

saauvine (approved):

PR #453, opened 2024-05-08
Force router advertisements leak

TIISR:
Hardcoding br-lan in radvd configuration to force Router Advertisements to leak to the whole mesh i.e. make all external devices self-configure with all the SLAAC prefixes, for example in a Laptop1 - CSL1 - CSL2 - Laptop2 scenario, Laptop1 and Laptop2 would both SLAAC with the prefixes from both CSL1 and CSL2.


johan-juntunen (approved):

joenpera (approved):

METII99 (approved):

saauvine (approved):

pentestiing (approved):
:adhesive_bandage:

PR #454, opened 2024-05-09
Mess bridge hotfix

TIISR:
Traffic multiplication appears to be due to a regression in the Lower/Upper Mess Bridge's (lmb/umb) even though they appear to be correctly configured in terms of forwarding database (bridge fdb) and port isolation (ip -d link show) so here is a bruteforce hotfix just dropping any traffic that should not be happening anyway.

WARNING: untested with more than 2 devices (which is required to be able to reproduced the issue) as I do not have access to more than 2 right now, but FWIW this hotfix does not seem to break anything at least between 2 devices.

PR #455, opened 2024-05-10
Mess bridge hotfix2

TIISR:
(Redoing the PR notably because I mistakenly created https://github.com/tiiuae/mesh_com/pull/454 as a branch of a previous PR branch and don't want to risk messing with rebasing etc)

Traffic multiplication appears to be due to a regression in the Lower/Upper Mess Bridge's (lmb/umb) even though they appear to be correctly configured in terms of forwarding database (bridge fdb) and port isolation (ip -d link show) so here is a bruteforce hotfix just dropping any traffic that should not be happening anyway.

WARNING: untested with more than 2 devices (which is required to be able to reproduced the issue) as I do not have access to more than 2 right now, but FWIW this hotfix does not seem to break anything at least between 2 devices.


pemanty (approved):
Tested against reported issue. Working fine.

joenpera (approved):

PR #456, opened 2024-05-13
Change Halow MTU handling

joenpera:
Now halow MTU is tried to set instead of printing error.

Jira-Id: SECO-5358

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
@@ -599,13 +599,9 @@ class CBMAAdaptation(object):
         self.logger.info("Setting MTU size for %s to %s", interface_name, mtu_size)
 
         try:
-            ret = subprocess.run(
+            subprocess.run(
                 ["ip", "link", "set", "dev", interface_name, "mtu", mtu_size], check=True
             )
-            if ret.returncode != 0:
-                self.logger.error(
-                    "Error setting MTU size for %s! Error: %s", interface_name, ret
-                )
         except subprocess.CalledProcessError as e:
             self.logger.error(
                 "Error setting MTU size for %s! Error: %s", interface_name, e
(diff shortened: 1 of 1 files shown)


saauvine (approved):

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
```diff
@@ -605,10 +599,13 @@ def __set_mtu_size(self, interface_name: str, interface_color: str = None) -> No
         self.logger.info("Setting MTU size for %s to %s", interface_name, mtu_size)
 
         try:
-            subprocess.run(
-                ["ip", "link", "set", "dev", interface_name, "mtu", mtu_size],
-                check=True,
+            ret = subprocess.run(
+                ["ip", "link", "set", "dev", interface_name, "mtu", mtu_size], check=True
```
pentestiing, line 603: I believe the `check=True` shall be removed if the `returncode` will be checked below, otherwise an [CalledProcessError](https://docs.python.org/3.9/library/subprocess.html#subprocess.CalledProcessError) will be raised when the command returns non-zero :boom:

pentestiing (changes requested):
:volcano:

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
joenpera: [78001c8](https://github.com/tiiuae/mesh_com/pull/456/commits/78001c855c0500804ddeda6c0b3412b7a4207b9e)

saauvine (approved):

modules/sc-mesh-secure-deployment/src/nats/src/cbma_adaptation.py
pentestiing: Ah, in this case, if you are only handling the CalledProcessError and are not interested in the command output you can use subprocess.check_call() instead of subprocess.run() 🧄

pentestiing (approved):
There's a lil' optional improvement suggested above :tokyo_tower:

PR #457, opened 2024-05-15
Field Test Logger update to get halow support working

joenpera:
Jira-Id: SECO-5192

### DIFF
--- a/common/tools/field_test_logger/wifi_info.py
+++ b/common/tools/field_test_logger/wifi_info.py
@@ -11,6 +11,7 @@ class WifiInfo:
         self.__country = ""
         self.__txpower = ""
         self.__noise = ""
+        self.__snr = "NaN"         # halow specific
         self.__rx_throughput = 0
         self.__tx_throughput = 0
         self.__stations = {}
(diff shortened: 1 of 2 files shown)


common/tools/field_test_logger/wifi_info.py
```diff
@@ -23,6 +24,14 @@ def __init__(self, interval, interface, batman_interface):
 
     # ----------------------------------------
 
+    def __run_command(self, cmd):
+        try:
+            with subprocess.Popen(cmd, stdout=subprocess.PIPE) as proc:
+                return proc.communicate()[0].decode().rstrip()
```
pentestiing, line 30: Use `subprocess.run(cmd, text=True).stdout` instead (you might also want to capture stderr with `stderr=subprocess.STDOUT`), or use `subprocess.check_output(cmd, text=True)` and capture the `CalledProcessError` exception if the command shouldn't fail :construction:

pentestiing (approved):
Added an optional recommendation to use modern non-Popen `subprocess` functions :fondue:

common/tools/field_test_logger/wifi_info.py
joenpera: Yes, there is a lot of things to re-write.   Now we are trying to keep this process alive as a first 🩹

pemanty (approved):

juhamunikie (approved):

PR #465, opened 2024-07-01
Upgrade CBMA to v0.1.8: Support TLS 1.3 + MACsec config v1 + MTU fix

pentestiing:
This PR brings:
- Support for TLS 1.3 (TLS 1.2 is still transitionally supported) for RPI <=> iMX8 compatibility
- First implementation of the extended MACsec config versioning (doesn't add anything new but paves the way for future extensions)
- MTU fixes for the standalone and simulation scripts

And, as always, some little refactoring here and there :tada: 

Jira-ID: SECO-5921, SECO-5922, SECO-6162


modules/sc-mesh-secure-deployment/src/nats/cbma/secure_socket/client.py
```diff
@@ -41,7 +41,7 @@ def __init__(self,
         self.server_port = server_port
         self.macsec_callback = macsec_callback
 
-        super().__init__(certificates, SSL.TLSv1_2_METHOD)
+        super().__init__(certificates, SSL.SSLv23_METHOD)
```
johan-juntunen, line 44: I think this will work, although it has deprecation warning.

The newer way is to pass SSL.TLS_CLIENT_METHOD and set minimum version by ctx options.

modules/sc-mesh-secure-deployment/src/nats/cbma/secure_socket/server.py
```diff
@@ -32,7 +32,7 @@ def __init__(self,
         self.client_ipv6 = client_ipv6
         self.macsec_callback = macsec_callback
 
-        super().__init__(certificates, SSL.TLSv1_2_METHOD)
+        super().__init__(certificates, SSL.SSLv23_METHOD)
```
johan-juntunen, line 35: I think this will work, although it has deprecation warning.

The newer way is to pass SSL.TLS_SERVER_METHOD and set minimum version by ctx options.

johan-juntunen (approved):
Looks good to me

modules/sc-mesh-secure-deployment/src/nats/cbma/secure_socket/client.py
pentestiing: Yes, that's right, but this method and the `set_min_proto_version` are only available from PyOpenSSL 21 and we only have 19.1 in our comms devices.

For completeness, CBMA `old-requirements-fixes` branch is the one being pushed to mesh_com with this deprecated code, but `develop` branch already use the new code:
- https://github.com/tiiuae/cbma/blob/develop/secure_socket/client.py#L44
- https://github.com/tiiuae/cbma/blob/develop/secure_socket/secure_context.py#L46

:star2:

modules/sc-mesh-secure-deployment/src/nats/cbma/secure_socket/server.py
pentestiing: Same as https://github.com/tiiuae/mesh_com/pull/465#discussion_r1661874328

TIISR (approved):

PR #466, opened 2024-07-11
MDM Agent: Handle NetlinkDumpInterrupted except in ip.get_links()

pentestiing:
The title :up: 

This fixes an exception (`NLM_F_DUMP_INTR`) raised by the kernel that happens in some cases when an interface is not yet ready.

According to the official documentation the fix is to try perform the operation again :raccoon: 

Relevant links:
- https://docs.kernel.org/core-api/netlink.html#support-dump-consistency
- https://docs.kernel.org/next/userspace-api/netlink/intro.html#dump-consistency

Jira-ID: SECO-6417


johan-juntunen (approved):
Looks good to me

METII99 (approved):

TIISR (approved):

PR #468, opened 2024-07-12
Fix old pyroute error + increase radio setup timeouts in MDM Agent

pentestiing:
The title :crown: 

In a previous PR, handling `NLM_F_DUMP_INTR` via `NetlinkDumpInterrupted` in pyroute2 was introduced, but this error handler is not available in pyroute2 0.5.7, so we are replacing it with the generic `NetlinkError`.

Also there seems to be some issue in not having the `halow` interface ready in CMs, so the timeout for this check has been increased to hopefully workaround this issue

:cd:


METII99 (changes requested):
Rpi /opt/nats/src/ files missing, please add

pentestiing:
> Rpi /opt/nats/src/ files missing, please add

For future readers `/opt/nats` is automatically created on boot by some magical scripts, no need to change anything here :dragon:

TIISR (approved):

PR #469, opened 2024-07-15
MDM Agent: Upgrade hardcoded fallback mDNS IPv6 to <prefix>::100

pentestiing:
The title :arrow_up: 

Change needed after upgrading the MDM Servers to use VRRP :mushroom: 

Edit:

Bit more context, if the mDNS resolution fails the fallback IPv6 will be used to resolve `defaultmdm.local` but then the certificate won't match, thus resulting in the following error:
> 2024-07-11 09:50:21,415 :: comms.controller.mdm_agent :: ERROR    :: HTTP request failed with error: HTTPSConnectionPool(host='fdbb:1ef7:9d6f:e05d::1', port=5000): Max retries exceeded with url: /public/get_device_config/mesh_conf?device_id=04f021bf893310000000744cf195 (Caused by SSLError(SSLCertVerificationError("hostname 'fdbb:1ef7:9d6f:e05d::1' doesn't match 'defaultmdm.local'")))

### DIFF
--- a/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
+++ b/modules/sc-mesh-secure-deployment/src/nats/mdm_agent.py
@@ -80,7 +80,7 @@ class MdmAgent:
             self.__lock
         )
         self.__url: str = "defaultmdm.local:5000"  # mDNS callback updates this one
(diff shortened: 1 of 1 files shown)


saauvine (approved):

johan-juntunen (approved):

TIISR (approved):
100% performance increase!

PR #471, opened 2024-08-05
Updating MPTCP configurations to integrate with SLAAC

Monika-Research:
MPTCP configurations are changed inorder to use SLAAC IPv6 addresses.

Note: S90mptcp service is called only after checking if the interfaces are up in entrypoint_nats.sh. As the radios are not brought up at entrypoint_nats.sh, S90mptcp service is not run by default. As there is no recent plan of having MPTCP in meshshield, S90mptcp service is called manually for testing purpose.


joenpera:
What is the status with this one?  Should we keep this open for trials or merged to develop?

mbaddeley:
Let's close this and keep in trials just as non-integrated feature branch. We do not want this in develop as it's not an official release that will be properly tested.

PR #475, opened 2024-09-02
SECO-7217: halow should be black interface for multi-radio routing to…

murugesan-kumar:
"If Halow is designated as the white interface and WiFi as the black interface, the Batman V routing with expected throughput will not function smoothly. This is because WiFi's expected throughput will always be 1 (at bat1) due to the stacked interfaces (lower + upper CBMA), while Halow will provide the actual throughput. As a result, traffic will consistently be routed through the Halow interface. To address this, we need to designate Halow as a black interface, similar to WiFi."


KavinKumarTN (approved):
Changes are fine in Multi-radio context.

TIISR (approved):
Makes sense :)

joenpera (dismissed):
wrong target branch

murugesan-kumar:
> wrong target branch

changed to develop. thank you

TIISR (approved):

saauvine (approved):

murugesan-kumar:
> wrong target branch
changed to develop

joenpera (approved):

PR #476, opened 2024-09-11
SECO-7322: mesh rssi threshold for halow to -105dbm

murugesan-kumar:
halow range is not good due to default lower mesh rssi threshold. Applying -105dbm as mesh rssi threshold


umamaheshrouthu (approved):

pemanty (approved):

KavinKumarTN (approved):
Mesh RSSI Thresold to -105 required for HaLow long range performance.

PR #481, opened 2024-10-01
Fix typo that caused CBMA not to find upper CBMA private key path

johan-juntunen:
-typo caused enum instead of string to be passed as key path
-SECO-7892


saauvine (approved):
Good catch! This was already fixed in MS3.0 repo.

METII99 (approved):

PR #483, opened 2024-10-15
Update HaLow Guard Interval for better throughput 

murugesan-kumar:
Update HaLow Guard Interval to short for better throughput.
Verification completed near TII Masdar area.

### DIFF
--- a/common/scripts/mesh-11s_nats.sh
+++ b/common/scripts/mesh-11s_nats.sh
@@ -177,7 +177,7 @@ EOF
       # Radio parameters
       echo "set radio parameters"
       # /usr/local/bin/cli_app set txpwr fixed 23
-      /usr/local/bin/cli_app set gi long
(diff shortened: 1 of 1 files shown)


joenpera (approved):

common/scripts/mesh-11s_nats.sh
```diff
@@ -177,7 +177,7 @@ EOF
       # Radio parameters
       echo "set radio parameters"
       # /usr/local/bin/cli_app set txpwr fixed 23
-      /usr/local/bin/cli_app set gi long
+      /usr/local/bin/cli_app set gi short
```
KavinKumarTN, line 180: Minor: This short GI is only for HaLow or even Wi-Fi profile also?

KavinKumarTN (approved):
Short Guard Interval changes are fine, we have done experiment to verify these changes as well.

saauvine (approved):

